PR #619, opened 2021-11-24
Add Color Card Problem Component

jmski:
The Color card problem will change colors based on mouse clicks.
The cards initially start as grey, then it loops between green, purple and yellow in that order.
The color corresponds with a card type debit, credit or both. Which card type is displayed on the top left of the card.
The answer to the question is shown as red text.

The page still needs answer validation functions
Screen Shot 2021-11-24 at 2 47 18 PM


pages/credit-card.tsx
```diff
@@ -13,6 +14,7 @@ export default function CreditCard(props) {
         DCWordProblem, //debit card word problem
         CCDCMulti, //multiple choice
         CardColorProblem, //credit debit both color problem
+        TestComponent,
```
vithushan19, line 17: lets not have this extra stage. I think this assignment is good enough based on what is in the teacher's PDF. I would focus on completing those requirements before adding in new stuff. 

If you're going to add new screens, I'd want to see you draw it out on a piece of paper before attempting to build it. I used to do this all the time as a junior dev, but it's really easy to immediately want to try building something without knowing exactly how it's going to work. 

If you want to add your own stuff to this lesson, draw out what it'll look like first. Otherwise just stick to what the teacher has in her assignment

components/credit-card/CreditDebitInfo.tsx
```diff
@@ -23,25 +49,43 @@ const CreditDebitInfo = () => {
...
+                        </div>
+                        ))}
+                    </div>
+                </div>
+
+                <div className="pl-6">
+                    <h2 className="text-2xl mb-4">How are they the same?</h2>
+                    <div className=" bg-gray-200 p-4">
+                        <ul className="flex flex-col gap-4">
+                            <li>Allow a person to make {whiteSpace()}</li>
```
vithushan19, line 81: I'm not down to merge this in as is. Some ideas:
- change this to be a text input that a student has to type in
- change this to a dropdown where the student has to select from a list of options
- give me some time to explore a re-usable drag and drop component

components/credit-card/CreditDebitInfoData.tsx
```diff
@@ -0,0 +1,50 @@
+import { WordBank } from "../../pages/api/credit-card/WordBank";
+
+export const WordBankData: WordBank[] = [
```
vithushan19, line 3: variable names should not be capitalized. Rename this to `wordBankData`

components/credit-card/DCWordProblem.tsx
```diff
@@ -3,30 +3,29 @@ import React from 'react'
 const DCWordProblem = () => {
```
vithushan19, line 3: can you rename this to DebitCardWordProblem?

components/credit-card/DragDrop.tsx
```diff
@@ -0,0 +1,98 @@
+import React, { Component } from 'react';
+
+export default class DragDrop extends Component {
```
vithushan19, line 3: can we remove this file for now, until we figure out drag and drop

components/credit-card/FinanceCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export interface FinanceCardProps {
+    question: string,
+    answer: string,
+}
+
+const FinanceCard = ({question, answer}: FinanceCardProps) => {
+    const [color, setColor ] = useState("bg-gray-300")
+    const [borderColor, setBorderColor ] = useState("border-gray-500")
+    const [cardTitle, setcardTitle ] = useState("Click to Change")
```
vithushan19, line 11: c should be capitalized in `setCardTitle`

components/credit-card/FinanceCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+    
+    const onColorChangeRequested = () => {
+        if (color == "bg-gray-300") {
```
vithushan19, line 14: you could add an OR operator to this condition and remove the last else if case.

ie `color === 'bg-gray-300' || color === 'bg-yellow-300'`

components/credit-card/FinanceCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+            setColor("bg-green-300")
+            setBorderColor("border-green-500")
+            setcardTitle("debit")
+
+        } else if (color == "bg-green-300") {
```
vithushan19, line 19: you should use === instead of ==

vithushan19: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness

pages/api/credit-card/WordBank.ts
```diff
@@ -0,0 +1,5 @@
+export interface WordBank {
```
vithushan19, line 1: do we need this? what's this for?

vithushan19: better to have the type in the file that's used in

vithushan19 (commented):
nice, looking good. I'll look into a drag and drop component. Consider some of my other suggestions if you want to move forward

pages/api/credit-card/WordBank.ts
jmski: It was used in the Drag n' Drop component so i'll remove this

components/credit-card/CreditDebitInfo.tsx
jmski: changed to a dropdown with 2 options. Having issues trying to iterate the options from an array

vithushan19 (approved):
Looks good to me, but doesn't look like you addressed any of the code style suggestions. Ask Ajevan or Mayu for help on iterating an array of options. But your approach should work for now.

PR #622, opened 2021-11-30
Create Redux State Manager for War Math Game 

Ajevan:
Create Redux State Manager (including creating slice file, edits to the reducer file to include the War Game slice, as well as test cases for each action within the slice file).


redux/warGameSlice.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    playerTwoReady: true,
+  });
+});
+
+test("test both players ready and start round action", () => {
+  const firstState = reducer(initialState, setPlayerReady(1));
+  const secondState = reducer(firstState, setPlayerReady(2));
+  expect(reducer(secondState, startRound(null))).toEqual({
+    cardListPlayerOne: [
+      { answer: 0, question: "" },
```
vithushan19, line 64: if we are starting a round and assigning values for playerOneCurrentCard and playerTwoCurrentCard, should we remove those cards from the respective card lists? Or should we keep them there for now and remove them later. Remember at the end of the round, the winning player needs to take both cards and put it at the bottom of their card list.

redux/warGameSlice.test.ts
```diff
@@ -0,0 +1,118 @@
...
+      question: "",
+    },
+    playerTwoCurrentCard: {
+      answer: 3,
+      question: "",
+    },
+  });
+});
+
+// test("should handle a todo being added to an empty list", () => {
```
vithushan19, line 87: remove these please

vithushan19 (approved):
Can you write a more professional description in your PR? And have a better PR title? First word should be a present tense verb

vithushan19:
Great job with this. You picked it up quickly on how to set up your own redux state. Really glad you found that video ahead of time, it made it a lot easier to discuss.

redux/warGameSlice.test.ts
Ajevan: I believe we should keep them there for now and then have another function to remove once they have been assigned and evaluated, just for having them in separate functions

redux/warGameSlice.test.ts
Ajevan: Removed in latest commit

PR #624, opened 2021-12-01
Create Redux tests for if either player one or two have won the round

Ajevan:
Create Redux tests for if either player one or two have won the round


redux/warGameSlice.ts
```diff
@@ -53,12 +59,43 @@ export const warGameSlice: Slice = createSlice({
         state.playerTwoCurrentCard = state.cardListPlayerTwo[0];
       }
     },
-    //Action that handles winning
+    //Action that handles winning and also handles moving cards from other player and from current card they chose
+    ifPlayerWon: (state: WarGameState, action: PayloadAction<number>) => {
```
vithushan19, line 63: can we rename this action? the other ones all start with a present tense verb like set or start.

how about something like finishRound? or scoreRound?

vithushan19 (approved):
awesome work!

redux/warGameSlice.ts
Ajevan: Renamed in newer commit

PR #627, opened 2021-12-03
Create Redux Reset State Function and test case

Ajevan:
Create Redux Reset State Function and test case

### DIFF
--- a/redux/warGameSlice.test.ts
+++ b/redux/warGameSlice.test.ts
@@ -284,7 +284,7 @@ test("test reset state", () => {
   const eleventhState = reducer(tenthState, startRound(null));
   const twelvethState = reducer(eleventhState, finishRound(1));
   const thirteenthState = reducer(twelvethState, finishRound(1));
-  expect(reducer(thirteenthState, resetRound("RESET"))).toEqual({
+  expect(reducer(thirteenthState, resetRound(null))).toEqual({
     cardListPlayerOne: [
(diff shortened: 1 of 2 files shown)


redux/warGameSlice.test.ts
```diff
@@ -268,3 +269,41 @@ test("test if game is over", () => {
...
+  const fifthState = reducer(fourthState, setPlayerReady(1));
+  const sixthState = reducer(fifthState, setPlayerReady(2));
+  const seventhState = reducer(sixthState, startRound(null));
+  const eighthState = reducer(seventhState, finishRound(1));
+  const ninthState = reducer(eighthState, setPlayerReady(1));
+  const tenthState = reducer(ninthState, setPlayerReady(2));
+  const eleventhState = reducer(tenthState, startRound(null));
+  const twelvethState = reducer(eleventhState, finishRound(1));
+  const thirteenthState = reducer(twelvethState, finishRound(1));
+  expect(reducer(thirteenthState, resetRound("RESET"))).toEqual({
```
vithushan19, line 287: why do you have to pass this string into resetRound? why can't we just pass in null?

vithushan19 (approved):

redux/warGameSlice.test.ts
Ajevan: We can, I thought to do a condition for reset, but for now we can also pass in Null, will make this change in new commit

PR #642, opened 2022-01-14
Refactor MC Quiz for JS and CSS

LavaBoy202:
This PR does:
- Refactors JS and CSS quiz components

### DIFF
--- a/components/coding/MCQuiz.tsx
+++ b/components/coding/MCQuiz.tsx
@@ -1,0 +1,98 @@
+import React, { useState } from "react";
+import { GuessData } from "../../pages/api/guessData";
+import {
+  MultipleChoiceSentence,
+  MultipleChoiceSentenceProp,
+} from "../questionTypes/MultipleChoiceSentence";
+
+export interface MCQuiz {
+  questionData: MultipleChoiceSentenceProp[];
+}
+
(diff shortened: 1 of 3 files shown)


components/coding/MCQuiz.tsx
```diff
@@ -0,0 +1,98 @@
...
+        className={`rounded-xl shadow-xl mt-4 p-4 transform transition-all bg-blue-100 ${
+          isShaking ? "animate-shake" : ""
+        } `}
+        onAnimationEnd={() => {
+          setIsShaking(false);
+        }}
+      >
+        <div>
+          {notComplete ? (
+            questionData[index].image ? (
```
LavaBoy202, line 41: Is using nested Ternary operators good practice?

vithushan19: yeah its fine, if its formatted properly

vithushan19, line 41: I feel like this ternary isn't needed though? Can you just pass down `questionData[index].image` no matter what. And sometimes it'll just be undefined and in that case the image just won't render?

vithushan19 (approved):
looks good to me. one suggestion about possibly removing one of the ternary's

PR #650, opened 2022-02-03
Javascript Assignment Page

LavaBoy202:
This PR:
- Creates a Webpage on instructions on how to complete the Javascript 
Screen Shot 2022-02-03 at 2 00 28 PM
Screen Shot 2022-02-03 at 2 00 46 PM


public/coding/JavaScript/PokemonTemp.js
```diff
@@ -0,0 +1,307 @@
+const PokemonData = {
```
vithushan19, line 1: It's overwhelming to have this pokemon data and the getFireblastFunction in the same file. Can we split them up into two different files? 

Eventually you are going to build a Coding Challenge UI that will look like this: https://codesandbox.io/s/holy-silence-oszdx?file=/src/App.js

So we will want to be able to support multiple file for a single coding challenge.

vithushan19 (approved):

PR #662, opened 2022-03-08
Sidebar Component

LavaBoy202:
This PR does:
- Creates the Sidebar Component

Screen Shot 2022-03-08 at 4 39 44 PM


components/coding/studentPortal/Sidebar.tsx
```diff
@@ -16,97 +16,95 @@ export const Sidebar: React.FC<SidebarProps> = ({}: SidebarProps) => {
...
+                    : "/images/profileInactive.svg"
+                }
+              />
+              Profile
             </div>
           </div>
         </div>
+        <div className="flex flex-col justify-end mb-18">
+          <div className="flex flex-wrap ">
+            <img className="w-7 h-7 mr-3" src="/images/logoutButton.svg" />
```
vithushan19, line 104: stick to multiples of 4 as best you can for tailwind units. ie use w-8 instead of w-7, and mr-4 instead of mr-3

components/coding/studentPortal/Sidebar.tsx
```diff
@@ -16,97 +16,95 @@ export const Sidebar: React.FC<SidebarProps> = ({}: SidebarProps) => {
...
+            className={`p-4 border-l-4 ${
+              isStageActive("Profile")
+                ? "border-charmander text-charmander"
+                : ""
+            }`}
+            onClick={() => onStageClick("Profile")}
+          >
+            <div className="flex flex-wrap">
+              <img
+                className="w-7 h-7 mr-3"
```
vithushan19, line 91: use 8 and 4

vithushan19:
I wouldn't use the stage pattern for this component. If you use the stage pattern and the user clicks around, their never going to change url pages. If they click around different pages, and click back it's going to do weird stuff. 

Instead I think that we should just have a separate page for Dashboard, Classroom, and Profile. Each page will use the same Sidebar navigation component. 

But clicking on a nav menu link should go to a new page, not set a state variable.

Let me know if you want to discuss this. I'm suggesting you remove the state variable for stages, and the NextJS <Link> tag instead. You can have both the classrooms and profile links to navigate to the old profile page.

vithushan19 (approved):

PR #664, opened 2022-03-09
Progress Bar Component

LavaBoy202:
This PR does:
-Creates a progress bar
-Requires the prop completed %
Screen Shot 2022-03-09 at 3 17 05 PM


components/coding/studentPortal/ProgressBar.tsx
```diff
@@ -0,0 +1,26 @@
...
+  completed,
+}: ProgressBarProps) => {
+  const fillerStyles = {
+    height: "100%",
+    width: `${completed}%`,
+    borderRadius: "inherit",
+  };
+  return (
+    <div>
+      <div className="h-8 w-full bg-gray-300 m-32 border rounded-full">
```
vithushan19, line 17: im not sure that you want to include the margin here. You may want put it in the parent instead, so the progress comes with no margin, but the parent that wants to plug it in can provide the right amount of margin

vithushan19 (approved):
nice and clean components! I love it.

PR #667, opened 2022-03-15
CSS Hi-fi design implementation

LavaBoy202:
This PR does:
- Implements the hi-fi design for the css page
- Adds a green colour called turtwig to the tailwinds config
Screen Shot 2022-03-15 at 2 04 24 PM

### DIFF
--- a/pages/course/[courseId]/unit/CSS/1.tsx
+++ b/pages/course/[courseId]/unit/CSS/1.tsx
@@ -4,6 +4,9 @@ import { Carousel } from "../../../../../components/coding/Carousel";
 import CSSQuiz from "../../../../../components/coding/CSSQuiz";
 import Navbar from "../../../../../components/ui/Navbar";
 import { MultipleChoiceSentence } from "../../../../../components/questionTypes/MultipleChoiceSentence";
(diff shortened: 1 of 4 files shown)


pages/course/[courseId]/unit/CSS/1.tsx
```diff
@@ -45,90 +48,113 @@ const CSS = () => {
...
-              {assignments.map((it) => (
-                <div>
-                  <a className="text-blue-600 underline" href={it.link}>
-                    {it.title}
-                  </a>
-                  <iframe
-                    src={it.link}
-                    className="w-full p-4 bg-white border-4 border-purple-400 shadow-lg h-96"
-                  />
+              {false && (
```
Comment (vithushan19) [inline pages/course/[courseId]/unit/CSS/1.tsx:85]: Why are you doing this?

vithushan19 (approved):
Looks good, but why are you hiding certain divs?

We'll have to go thru the site to make it look good on mobile later. Let's sync on that this week.

pages/course/[courseId]/unit/CSS/1.tsx
Comment (LavaBoy202) [reply pages/course/[courseId]/unit/CSS/1.tsx:85]: I just did this so I can use the code again for other pages

PR #691, opened 2022-04-19
Submission Confirmation to Complete Assignment Nodes

LavaBoy202:
This PR does:
- creates an submission confirmation page to complete Assignment Nodes


vithushan19 (approved):
nice work!

PR #704, opened 2022-06-21
changes to NBA sports assignment instructions page

iamnotreddy:
High Level Changes:
- replace old instructions page with new instructions provided by vithushan
- add async function / state var to fetch actual player JSON data, which will be used for examples on real data... added an example for GSW warriors players :D 
- make each challenge collapsable using conditional render on "visible" state var for each challenge and Skillify button component

def need to re-style the buttons so they look smaller, more uniform on the page... I started by using the skillify button component but maybe will switch to a custom icon, but lmk what you think

UI Changes:
Screen Shot 2022-06-21 at 12 36 20 PM


pages/studentPortal/intro/Javascript/sports/instructions.tsx
```diff
@@ -1,77 +1,184 @@
...
           </div>
-          <div>
-            <p className="text-charmander">Step 4:</p>
+
+          {challengeThreeVisible && (
             <p>
-              We must now figure out a way to fiter only the plays of etype of
-              15.
+              Write a function that sorts this array based on your rating
+              function and print out the top 10 players. // (hint. use map() to
```
vithushan19, line 138: can you make the hint hidden by default, and have a show hint button that reveals the hint?

vithushan19 (commented):
looks great so far!

vithushan19 (approved):

PR #707, opened 2022-06-24
create user goals table in DB, render on profile

iamnotreddy:
- created a user_goals table in Hasura - lmk what you think about the schema
- added fetchUserGoals gql query, types to define query responses
- add fetchUserGoals useQuery hook to profile, userGoals state variable which is set on useQuery complete
- render goals from userGoals state var in place of static goals with some conditional rendering logic to prevent page load failures

Screen Shot 2022-06-24 at 11 31 44 AM

### DIFF
--- a/components/coding/GoalsSection.tsx
+++ b/components/coding/GoalsSection.tsx
@@ -1,0 +1,56 @@
+import { useQuery } from "@apollo/client";
+import React, { useState } from "react";
+import {
+  FetchUserGoalsDataResponse,
+  FETCH_USER_GOALS,
+  UserGoalsData,
+} from "../../graphql/fetchUserGoals";
+import { useAuth } from "../../lib/authContext";
+
+export type GoalsSectionProps = {
+  user: any;
+};
+
+export default function GoalsSection({ user }: GoalsSectionProps) {
+  const [userGoals, setUserGoals] = useState<UserGoalsData[]>([]);
+  const { loading: userGoalsLoading } = useQuery<FetchUserGoalsDataResponse>(
+    FETCH_USER_GOALS,
+    {
+      variables: {
+        userId: user.uid,
+      },
+
+      onCompleted: (data: FetchUserGoalsDataResponse) => {
+        setUserGoals(
+          data.user_goals.map((it) => {
+            return {
+              __typename: it.__typename,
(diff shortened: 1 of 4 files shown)


pages/profile.tsx
```diff
@@ -27,12 +31,37 @@ export default function Profile(props) {
           profileImage: data.users[0].profile_image,
         });
       },
+    });
+
+  const { loading: userGoalsLoading } = useQuery<FetchUserGoalsDataResponse>(
```
vithushan19, line 36: you should move all the Goals code into a new component called `GoalsSection` move the query into that component as well as the UI for it. 

That way we only query the API when the GoalsSection component is rendered. If you keep the query in the profile page, we always fetch the users goals, even in cases where we want to hide the section by default.

Moving the query logic into a GoalsSection component, allows us to re-use the component in different parts of the student portal as well.

iamnotreddy:
screenshot after "componentizing" profile and goals section:
Screen Shot 2022-06-24 at 12 44 11 PM

components/coding/GoalsSection.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  return (
+    <>
+      {userGoalsLoading ? (
+        <div>Loading...</div>
+      ) : (
+        userGoals.map((it) => {
+          return (
+            <div className="text-white bg-murkrow text-center rounded-full mx-5 py-2 mb-5">
+              {it.goalName}
```
vithushan19, line 48: Amazing

pages/profile.tsx
```diff
@@ -1,83 +1,23 @@
...
       <h2 className="mt-14 mb-9 font-bold text-lg">Goals</h2>
+
       <div className="grid grid-cols-1 sm:grid-cols-3 mb-16">
-        <div className="text-white bg-murkrow text-center rounded-full mx-5 py-2 mb-5">
-          Get A Tech Job
-        </div>
-        <div className="text-white bg-murkrow text-center rounded-full mx-5 py-2 mb-5">
-          Learn A New Skill
-        </div>
+        <GoalsSection user={user} />
```
vithushan19, line 18: Amazing!

vithushan19 (approved):

PR #709, opened 2022-06-28
fetch user goals from hasura on goals page

iamnotreddy:
goals page now fetches data from user_goals table in hasura:
- add FETCH_USER_GOALS query, useQuery hook and store data into userGoals state var to goals page
- add GoalsSection component that takes in two props: filtered array of the userGoals response based on section, and a header containing the section name and relevant heroicons
- use goalsSections var to define props for each section that will be rendered; map through this and render each GoalComponent

Other changes:
- added new fields to user_goals table in hasura and updated gql / types: isComplete, targetDate
- defined new types for the GoalsSectionComponent, moved to a helpers file in api folder
- renamed some components and types from the user profile page's Goals section to help differentiate between the new "Goals Section" on the user goals page

Screenshot (reading from hasura table for my Skillify user ID):
Screen Shot 2022-06-28 at 1 54 45 PM


pages/goals.tsx
```diff
@@ -6,132 +6,84 @@ import {
...
-          createdDate: "2022-06-01",
-          targetDate: "2022-09-01",
-        },
-        {
-          goalName: "Learn SQL",
-          createdDate: "2022-06-01",
-          targetDate: "2022-09-01",
-        },
-      ],
+      rows: userGoals.filter((goal) => !goal.isComplete && !goal.isActive),
```
vithushan19, line 65: so when a goal goes from complete to archived, we have to mark it as incomplete? 

would it make more sense to add a new column called isArchived? And then only show a goal as completed if `isComplete && !isArchived`

pages/goals.tsx
```diff
@@ -6,132 +6,84 @@ import {
...
-              );
-            })}
-          </div>
-        );
-      })}
+      {userGoalsLoading ? (
+        <div>Loading...</div>
+      ) : (
+        <div>
+          {goalsSections.map((section) => {
```
vithushan19, line 75: amazing! so clean

pages/api/goals/goalsHelpers.ts
```diff
@@ -0,0 +1,12 @@
...
+import { UserGoalsData } from "../../../graphql/fetchUserGoals";
+
+export type GoalsSection = {
+  header: GoalsSectionHeader;
+  rows: UserGoalsData[];
+};
+
+export type GoalsSectionHeader = {
+  sectionName: string;
+  manageIcons: Array<ReactElement>;
```
vithushan19, line 11: is manageIcons really part of the Header? seems like it represents the icons you want to display for each row in this section. 

I might remove this type entirely, and move both `sectionName` and `manageIcons` into the `GoalsSection` type

vithushan19: not a big deal either way, im fine leaving it as is.

pages/api/goals/goalsHelpers.ts
```diff
@@ -0,0 +1,12 @@
+import { ReactElement } from "react";
+import { UserGoalsData } from "../../../graphql/fetchUserGoals";
+
```
vithushan19, line 3: helpers generally implies that this file will have helper functions. I would rename it types.ts if you need to keep it.

vithushan19: I'm also wondering if you need the GoalsSection type at all, since it's identical to the GoalsSectionProps type. I wonder if there's a benefit to having two different types rather than sharing the same one.

vithushan19 (commented):
Nice work fetching the data from Hasura. I have some questions around your types, and the need for the helpers file. Everything else looks great!

vithushan19 (approved):
Lgtm. Were you able to see my PR around adding a story for storybook?

PR #717, opened 2022-07-07
add goal page

iamnotreddy:
- created an add goal page where user can write a description of their goal, set a target completion page, and get routed back to the goals overview page to see their updated goal... page is similarly constructed to edit goals page, but only includes relevant fields for adding a new goal
- create "add goal" button on goals overview page which navigates to add goal page

screenshot:

Screen Shot 2022-07-07 at 8 02 01 AM

### DIFF
--- a/pages/goals/addGoal.tsx
+++ b/pages/goals/addGoal.tsx
@@ -11,6 +11,7 @@ const EditGoalsPage = () => {
 
   const router = useRouter();
 
+  // initialize only with values required by DB to add goal
   const [newGoalValues, setNewGoalValues] = useState({
     userId: user.uid,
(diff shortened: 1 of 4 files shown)


pages/goals/addGoal.tsx
```diff
@@ -0,0 +1,76 @@
...
+            setNewGoalValues((prevState) => ({
+              ...prevState,
+              targetDate: new Date(e.target.value),
+            }));
+          }}
+        />
+      </div>
+
+      <div>
+        <Button
```
vithushan19, line 61: Can we disable this button, if the goal text is empty?

PR #722, opened 2022-07-13
implement redux on sidebar

iamnotreddy:
remove activePage useState hook and replace with redux selector and dispatcher


redux/sidebarSlice.ts
```diff
@@ -0,0 +1,38 @@
...
+  | "profile"
+  | "labs"
+  | "goals";
+
+export interface SidebarState {
+  activePage: SidebarPage;
+}
+
+const initialState: SidebarState = {
+  activePage: "goals",
```
vithushan19, line 18: Should this be dashboard instead?

vithushan19 (approved):
Looks good to me

PR #725, opened 2022-07-21
implement redux into skillRankings

iamnotreddy:
** this PR isn't fully working yet, need some feedback / guidance ** 

changes in this PR:
- move skillRankings child components into separate folder
- add skillRankings redux slice, replace state variables in parent component with redux selectors
- dispatch transformed data to redux selector on useQuery onComplete, pass as prop into SkillSection child
- attempt to render from prop

I'm running into some issues with the order of operations. Confirmed I can read real data from the redux selector `skillRatings` but the render fails because the component tries to render the prop data before its loaded from the selector. When I just render the prop data through JSON.stringify(), it takes a few seconds to display on page load. The conditional rendering isn't addressing this


vithushan19:
Your conditional render is checking whether or not `skillRatings` is undefined or not, but it is always defined because you initialize it as an empty array. Your render is failing, because the component assumes that there is at least one skill in the array. So you when calling `skillSection[0].name` it fails because skillSection is an empty array. 

An empty array is not the same thing as being undefined. I was able to fix your error by adding an additional check to the conditional render. I check that not only is the skillSection defined, but also that it has a length greater than 0.

That should fix your issue.

iamnotreddy:
> 

🥲

iamnotreddy:
@vithushan19 - this PR is ready for review

I pushed commits to modify the updateSkillRatings action in the redux slice - can now successfully update the value from the sliders in the redux store. All of the state vars and callbacks have been replaced - the redux implementation is complete and appears to be working.

Can you take a look at the updateSkillRatings action? I tried to create a copy of the skillRatings selector, update the value with the payload, then incorporate that into the skillRatings selector. 

I originally tried to do something at the end along the lines of: `return {...state, skillRatingsCopy}` but kept running into this issue: `[Error: [Immer] An immer producer returned a new value *and* modified its draft...`

My current code works but I don't know why lol... I read it as I'm just updating the copy not the actual selector

redux/skillRatingsSlice.ts
```diff
@@ -30,10 +30,25 @@ export const skillRatingsSlice: Slice = createSlice({
...
     },
+    updateSkillRatings: (
+      state: SkillRatingsState,
+      action: PayloadAction<{ newStudentRating: Number; skillId: String }>
+    ) => {
+      if (action.type == "skillRatings/updateSkillRatings") {
+        const skillRatingsCopy = [...state.skillRatings];
+
+        const index = skillRatingsCopy.findIndex(
+          (obj) => obj.skillId == action.payload.skillId
```
iamnotreddy, line 39: 🤷

vithushan19 (commented):
Will take a look but I think you should add a unit test in this pr. It's good practice to add tests in the same pr as your change. Take a look at some of the other slice tests.

vithushan19:
Some companies have pr checks in place to prevent you from merging PRs unless a test is accompanied

vithushan19 (approved):

vithushan19 (approved):
amazing!

PR #730, opened 2022-08-15
Move coachCard from classroom folder to coaches folder

vithushan19:
We forgot to rename the folder from classroom to coaches when we created the coaches the page. This PR fixes that and renames the folder.


vithushan19 (commented):
LGTM

PR #733, opened 2022-08-16
added animal

kwiedenhaupt:
Added an animal to the game


vithushan19 (changes requested):
One thing that I failed to mention is that we are using tailwind v2 not v3. So we actually don't have access to the slate colours. Take a look at the v2 docs here: https://v2.tailwindcss.com/docs/text-color

And whenever you create a PR that changes the UI, it's best practice to include a screenshot of what the UI will look like. Can you add this image into your PR description?

Screen Shot 2022-08-16 at 12 46 02 PM


Also the image looks

kwiedenhaupt:
goatGitHub

kwiedenhaupt:
This is a preview of what the UI will look like.

vithushan19 (approved):

PR #735, opened 2022-08-17
Create components for Workshops Page and add type

Sukeesan:
Updated workshops page - added white card and bolded video headings.


vithushan19 (changes requested):
Can you create a React component called `<WorkshopCard/>` that takes in Props and reduce the duplication of code.

Also, can you change the title of your PR to be more descriptive.

vithushan19:
Video Feedback 

[Alt text](https://www.loom.com/share/e29914026e7a4c0aa7c996f3cc66b204)

components/coding/studentPortal/WorkshopsCard.tsx
```diff
@@ -0,0 +1,19 @@
+import { Button } from "../../ui/Button";
+
+// type WorkshopsCardProps = {
```
vithushan19, line 3: This is the right idea, take a look at other components and see how I defined the type for props.

You should uncomment this and add more properties

components/coding/studentPortal/WorkshopsCard.tsx
```diff
@@ -0,0 +1,19 @@
...
+//   title: string;
+// };
```
vithushan19, line 5: You should add additional properties for video and starterFiles

Video should be a string
And starterFiles should be an optional string

vithushan19: https://www.typescriptlang.org/docs/handbook/2/objects.html

Read the part called optional parameters

components/coding/studentPortal/WorkshopsCard.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export default function WorkshopsCard(props) {
+  return (
+    <div className="bg-white p-6 rounded-md shadow-md">
+      <h2 className="font-bold">{props.title}</h2>
+      <video src={props.video} controls={true} />
+      <a href={props.starterFile}>
+        <div className="mt-4">
+          {props.starterFile ? <Button label={"Starter Files"} /> : null}
```
vithushan19, line 14: Nice work!

pages/workshops.tsx
```diff
@@ -11,53 +11,39 @@ import { useQuery } from "@apollo/client";
...
-            />
-          </div>
-          <div>
-            <h2>Pokedex - July 21, 2022</h2>
-            <video
-              src="https://d3jppm0n5ndqu2.cloudfront.net/pokedex-july-21.mov"
-              controls={true}
-            />
-          </div>
+          <WorkshopsCard
```
vithushan19, line 26: You can clean up this code by creating an array of WorkshopsCardProps and putting all the values in that array.

Then you can map through that array to render a list of the card components.

What you have is correct but my suggestion is a better practice for cleaner code.

vithushan19: Look at this file for example

vithushan19 (changes requested):
Nice work! Just left a few comments to address

vithushan19 (approved):
Nice work! You have merge conflicts so we'll have to fix that before we merge. We can discuss that tomorrow

PR #736, opened 2022-08-17
adding useState to UnitPriceQuestion

kwiedenhaupt:
It returns with an error that submitGuess is undefined or is not a function..
Screen Shot 2022-08-17 at 9 10 06 PM
.


vithushan19 (approved):
looks great!

a few things that could have made this feature even better:
- randomizing your name instead of always saying `Kari`
- fixing the UI when the user enters an incorrect guess

Neither of these is blocking, so I'll approve and you can merge this PR in and move on to the next assignment.

PR #738, opened 2022-08-19
created component page and first row of blocks

kwiedenhaupt:
I was just playing around with the random numbers function. I cannot figure out yet how to get two random numbers chosen written out and then their product also chosen as well.

I also did not yet look into how to get the big board together.


vithushan19:
Let's just hard code the expression and numbers using an array for now. Let's not add the randomization in.

Just hardcode the strings.

components/math/multiplication.tsx
```diff
@@ -0,0 +1,29 @@
...
+export function getRndBlock() {
+  let x = randomize(1, 10);
+  let y = randomize(1, 10);
+  let z = x + y;
+  let blockArray = [x + " x " + y, z];
+  let block = getRandomItemFromArray(blockArray);
+  return block;
+}
+
+export default function MultiplicationGame() {
```
vithushan19, line 14: The filename and this function name for the component should match.

Maybe call it MultiplicationBlock or MultiplicationSquare?

components/math/multiplication.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { useEffect, useState } from "react";
+import { randomize } from "../../pages/api/questions/questionGenerator";
+import { getRandomItemFromArray } from "../../pages/api/random";
+
+export function getRndBlock() {
```
vithushan19, line 5: Let's not do the randomization now. Let's just create a component that's only job is to render a square given some string property

components/math/multiplication.tsx
```diff
@@ -0,0 +1,29 @@
...
+  let y = randomize(1, 10);
+  let z = x + y;
+  let blockArray = [x + " x " + y, z];
+  let block = getRandomItemFromArray(blockArray);
+  return block;
+}
+
+export default function MultiplicationGame() {
+  return (
+    <div>
```
vithushan19, line 16: Looks good. Looks like you're figuring out tailwind

pages/api/questions/questionGenerator.ts
```diff
@@ -130,6 +130,7 @@ export const generateQuestionForSkill = (
         unitPriceModel: {
           total: randomTotal,
           numberOfObjects: randomNumberOfObjects,
+          answer: "",
```
vithushan19, line 133: This looks like its not meant to be here. Maybe undo this? Or remove this change?

pages/studentPortal/labs/multiplication/game.tsx
```diff
@@ -0,0 +1,13 @@
+import MultiplicationGame from "../../../../components/math/multiplication";
+
+//create big box
+export default function BlockComponentGallery() {
```
vithushan19, line 4: This function name and the file name should match for best practices

vithushan19 (changes requested):
Focus on building the component first and defining the props for that component. We'll add the randomization later, for now you could even create the board where they all have the same expression and value. Just to break the problem down .

components/math/multiplication.tsx
kwiedenhaupt: I changed it the function and component file name to MultiplicationBlock

vithushan19 (approved):
Looks great. You can merge this in. The next step would be to create props to set the text inside a block. And a prop to determine if a block should be selected or not.

Then you can useState in the game component to keep track of which blocks should be selected.

We can sync on this on Monday

vithushan19:
You also have merge conflicts, which is an important thing for you to learn how to resolve.

A merge conflict happened because I you created a branch off of main, but then I merged in changes to some of the same files you're touching, before you got a chance to merge your PR in.

This happens all the time when you work on a team and multiple people are touching the same files.

You can try googling how to resolve a merge conflict or I can walk you through it on Monday.

Merge conflicts can get frustrating if you're unprepared.

So maybe work on the unit problem question follow up tasks until we get a chance to talk.

kwiedenhaupt:
That makes sense concerning the merge conflict.  I will wait to speak with you.  I have also been working on the unit price and I think I need more explanation of how the internal files are connected in order to make it work.  I just can't seem to figure it out lol.

PR #739, opened 2022-08-20
Kw add fruit unit price

kwiedenhaupt:
I was able to get the name to randomize pretty easily, but have tried for hours and can't figure out how to get the image and the name of the fruit to pull from the word problem question (fruitsMap) etc.

### DIFF
--- a/components/finance/UnitPriceQuestion.tsx
+++ b/components/finance/UnitPriceQuestion.tsx
@@ -65,7 +65,10 @@ const UnitPriceQuestion: React.FC<UnitPriceQuestionProps> = ({
                 type="number"
                 value={guess}
                 className="w-20 font-bold text-right border-2 border-gray-300"
-                onChange={(e) => setGuess(e.target.value)}
+                onChange={(e) => {
+                  e.stopPropogation();
+                  (e) => setGuess(e.target.value);
+                }}
               ></input>
             </p>
           </div>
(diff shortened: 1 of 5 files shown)


vithushan19:
https://github.com/skillify-ca/skillify-web/blob/main/pages/api/labs/questionGenerators/wordProblemQuestion.ts#L71

Could you reuse this function?

If you pass in the fruitsMap into that function it should randomly give you a fruit object.

The fruit object should have a name property and an image property.

Is that helpful?

components/finance/UnitPriceQuestion.tsx
```diff
@@ -1,19 +1,28 @@
...
 
 export interface UnitPriceQuestionProps {
   total: number;
   numberOfObjects: number;
   submitGuess: (guess: GuessData) => void;
   answer: string;
+  name: string;
 }
 
+const nameSelector = () => {
```
vithushan19, line 15: Instead of duplicating this function, could you add an export keyword to the existing function and import the function here?

kwiedenhaupt: Thank you! that makes much more sense on the names....still can't figure out the fruits. it's more involved than what I'm doing I'm sure.
Screen Shot 2022-08-20 at 10 21 45 PM

kwiedenhaupt:
Thank you for helping....I was totally stuck before....any quick suggestions on how to adjust the card?
Screen Shot 2022-08-20 at 10 36 22 PM

vithushan19:
Wooohoo!! awesome!

I would put the image to the right of the text input, that should give you enough space to show the submit button properly.

To put the text input and the image on the same row, I would wrap them both in a div and make that div a flexbox

vithushan19 (approved):
Great work, we don't include our a initials in the PR title as a best practice. 

PR titles should start with a present tense verb.

PR #747, opened 2022-08-25
created second player component

kwiedenhaupt:
added second player component to change squares to blue for the second player. 

I am working on the parent so I can add the child player 1 and player 2.  I made them separate components.  Am I on the right track?


components/math/MultiplicationBlock.tsx
```diff
@@ -4,22 +4,26 @@ export interface MultiplicationBlockProps {
   text: string;
   selected: boolean;
   onClick: boolean;
+  player: boolean;
 }
 
 export const MultiplicationBlock: React.FC<MultiplicationBlockProps> = ({
   text,
   onClick,
+  player,
```
vithushan19, line 13: You defined this new prop but you're not using it anywhere. If you read this prop inside your ternary statement you wouldn't need to create a duplicate blocks component.

components/math/MultiplicationBlock.tsx
```diff
@@ -4,22 +4,26 @@ export interface MultiplicationBlockProps {
...
   const handleSelected = () => {
     setSelected(!selected);
   };
+
   return (
     <div>
       <ul
         className={`w-20 h-20 flex justify-center items-center border-2 text-green-50 float-left
-        ${!selected ? "bg-green-400" : "bg-red-400"}`}
+        ${!selected ? "bg-green-400" : "bg-red-400"} `}
```
vithushan19, line 26: What if you had a nested ternary here? Could you determine the right colour based on the values of both selected and player?

components/math/MultiplicationBlockPlayer2.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { useEffect, useState } from "react";
+
+export interface MultiplicationBlockPlayer2Props {
```
vithushan19, line 3: Don't think we want this duplicate component. It's pretty much the same as your other one except it shows blue.

Could you use the player prop with your existing ternary to achieve the same thing?

vithushan19 (commented):
Think you have the right idea, but a duplicate block component just for changing the colour isn't the best plan. You can achieve that blue color background by modifying the ternary in your existing component

vithushan19 (approved):

pages/studentPortal/labs/multiplication/game.tsx
```diff
@@ -1,14 +1,114 @@
...
+
+  function handleReset() {
+    setGameState(initialGameState);
+  }
+  function longestSubarray(array, x) {
+    let maxlength = 0;
+    let sum = 0;
+    for (let i = 0; i < array.length; i++) {
+      let sum = 0;
+      if (array[i].winner == x) {
```
vithushan19, line 73: I'm not sure that winner is a good name here. Could we remove that property and if x == 1 then check if array[i].blockState == PLAYER_ONE_SELECTED

And if x==2 then check for PLAYER_TWO_SELECTED

pages/studentPortal/labs/multiplication/game.tsx
```diff
@@ -1,14 +1,114 @@
...
+        sum++;
+      } else {
```
vithushan19, line 75: Empty else case can be removed

vithushan19 (changes requested):

pages/studentPortal/labs/multiplication/game.tsx
kwiedenhaupt: It actually needed to contain let sum = 0 to reset the counter. I had it but must not have saved it lol So I added that.

pages/studentPortal/labs/multiplication/game.tsx
kwiedenhaupt: I'm confused as to what you mean.  The winner property was used to create an array that could be fed into the longestSubarray function in order to determine the winner.  As of now I have it logged into the console only but it does work.

pages/studentPortal/labs/multiplication/game.tsx
```diff
@@ -1,14 +1,114 @@
-import MultiplicationBlock from "../../../../components/math/MultiplicationBlock";
-//create big box
+import React, { useState } from "react";
+import MultiplicationBlock, {
+  BlockState,
+} from "../../../../components/math/MultiplicationBlock";
+import { Button } from "../../../../components/ui/Button";
+
 export default function BlockComponentGallery() {
+  const [player, setPlayer] = useState(false);
```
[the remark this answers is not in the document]
kwiedenhaupt: I changed it to isPlayerOneActive.

How would you do the second if you need to increment up and down? I'd love to see that.  so it could go from 1 to 2 back to 1 on a button click.

vithushan19 (commented):
This is looking really good! Let's discuss how to add unit tests tomorrow and how to implement the stage pattern!

vithushan19 (approved):
looks good to me! nice unit tests!

PR #750, opened 2022-08-30
working on layout for the grid of blocks

kwiedenhaupt:
working on making the grid layout work


kwiedenhaupt:
I am trying to work with the grid, but having trouble because I have to keep the InitialGameState blocks in order.  They do not layout like a frame snaking around but go in a grid starting over again at the next row.  All the blocks are being fed into one grid container due to the MultiplicationBlock component and I'm not quite getting place-content-between to work either...it does not push everything toward the edges.

vithushan19:
ping me for a review, once you merge these conflicts

PR #755, opened 2022-09-03
changes requested in the review

kwiedenhaupt:
I did the merge before the changes were made.  These are the changes.


vithushan19 (approved):
Looks good! I think you're close to moving on to databases and advanced react. You seem comfortable with typescript and useState.

For databases you're going to create an analytics dashboard that a teacher can use to monitor how their classroom is interacting with your game.

For advanced react you will be learning redux and will have to refactor your code to use redux.

After you do the redux, change the second player in your game to be an "AI" that randomly selects squares. When a player beats the AI they should move up a level and see 3 times multiplications. If they lose to the AI the student should go down a level.

A student's current level will need to be saved in a database and shown on the analytics dashboard.

We will discuss this in detail on Monday. For now focus on learning Redux toolkit and think about how your code needs to change to use Redux.

The redux folder in the codebase has lots of examples, but I think the quizSlice is the latest one that is easiest to read.

PR #756, opened 2022-09-04
Adding to GameBoard state

kavanparam:
- Randomized the board numbers (even numbers ranging from 4-24)
- Minor UI change

image

### DIFF
--- a/components/math/multiplicationConnect/GameBoard.tsx
+++ b/components/math/multiplicationConnect/GameBoard.tsx
@@ -4,7 +4,7 @@ const GameBoard = () => {
   return (
     <div className="px-28">
       <h2 className='text-xl pb-4'>GameBoard</h2>
-      <div className="grid grid-cols-5 border-t-2 border-l-2 border-t-blue-600 border-l-blue-600 bg-blue-900">
+      <div className="grid grid-cols-5 border-t-2 border-l-2 border-t-blue-600 border-l-blue-600 bg-blue-900 text-white">
         <div className='flex justify-center items-center border-r-2 border-b-2 h-28 border-b-blue-600 border-r-blue-600 bg-blue-900'>1</div>
         <div className='flex justify-center items-center border-r-2 border-b-2 h-28 border-b-blue-600 border-r-blue-600 bg-blue-900'>2</div>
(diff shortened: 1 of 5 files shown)


components/math/multiplicationConnect/GameBoard.tsx
```diff
@@ -0,0 +1,34 @@
+import React, { useState } from 'react'
+import { getRndInteger } from '../../../pages/api/random';
+
+const createGrid = () => {
+    let arr = [];
+    do{
+        let gridNumber = getRndInteger(4, 25);
+        (gridNumber%2===0) ? (
```
vithushan19, line 8: This can slow down your user experience greatly. If you keep getting odd numbers your user will be stuck in an infinite loop and it'll crash their browser.

Is there another way we can achieve this?

components/math/multiplicationConnect/GameBoard.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+
+const GameBoard = () => {
+    const [grid, setGrid] = useState(createGrid);
+
+    return (
+        <div className="px-28">
+            <h2 className='text-xl pb-4'>GameBoard</h2>            
+            <div className="grid grid-cols-5 border-t-2 border-l-2 border-t-[#149ECA] border-l-[#149ECA] text-white text-2xl">
+                {grid.map(i => <div className='flex justify-center items-center border-r-2 border-b-2 h-28 border-b-[#149ECA] border-r-[#149ECA] bg-blue-900'
```
vithushan19, line 27: Nice!

components/math/multiplicationConnect/GameBoard.tsx
vithushan19: We have a helper function called get random item from array. Why don't you create an array of all valid values and pick a random item from there?

vithushan19 (changes requested):

components/math/multiplicationConnect/GameBoard.tsx
kavanparam: I couldn't think of a workaround but this is a good idea, thanks!

vithushan19 (approved):
looks good to me!

PR #758, opened 2022-09-04
Merging completed work from branches

kavanparam:
- GameBoard now populated with random numbers according to game mechanics
- Dice now rolling two random numbers
- UI font size changes for readability

image


vithushan19 (approved):
looks good to me!

PR #764, opened 2022-09-06
Create new skill for algebra questions

vithushan19:
This PR defines a new skill called `solve for variable` and adds it to the Skills enum. I also had to define a new unit called `algebra` in our database.

Screen Shot 2022-09-06 at 4 46 10 PM


LuckyHariharan:
LGTM!

PR #772, opened 2022-09-15
Two player support

kavanparam:
Alternating player selection is now working, and game win can be determined between two players!

- removed `GameBoardBlock`'s `block` state & extracted logic to App level passing down function as prop (`blockClick()`)
- UI:
       * font changes for buttons
       * added shading to GameBoard for depth/separation in the UI
       * removed unnecessary heading  
       * added different colours for player selection (selected blocks & hover state for block to be selected).
- added `isPlayerOne` state to manage current player state that’s toggled inside `blockClick()`
- refactored `isSelected` property of `GameBoardBlock` to enum `selectedBy`
- refactored all setStates to make sure no state was modified directly — used previousState and creating copies of state in setState calls
- refactored `calculateWinner()` to accommodate checking game win between two players

image

PR #773, opened 2022-09-15
Merge two player support to main

kavanparam:
https://github.com/skillify-ca/skillify-web/pull/772


components/math/multiplicationConnect/GameBoard.tsx
```diff
@@ -3,30 +3,35 @@ import GameBoardBlock from "./GameBoardBlock";
...
   return (
-    <div className="px-28 pb-8">
-      <h2 className="text-xl pb-4">GameBoard</h2>
+    <div className="px-20 pb-10">
       <div className="grid grid-cols-5 border-t-2 border-l-2 rounded-2xl border-t-[#149ECA] border-l-[#149ECA] text-white text-2xl">
         {grid.map((blockData) => (
           <div
             key={blockData.id}
-            className={`border-r-2 border-b-2 h-28 border-b-[#149ECA] border-r-[#149ECA] bg-blue-900 
+            className={`border-r-2 border-b-2 md:h-32 sm:h-24 h-20 border-b-[#149ECA] border-r-[#149ECA] bg-blue-900
```
vithushan19, line 17: Awesome! Responsive breakpoints!

pages/studentPortal/labs/multiplication-connect/Index.tsx
```diff
@@ -1,31 +1,21 @@
...
-
-export const calculateWinner = (grid: GameBoardBlock[]) => {
-  // Algorithm ran on block click to determine win
+enum SelectedBy {
+  Unselected = "UNSELECTED",
+  PlayerOne = "PLAYERONE",
+  PlayerTwo = "PLAYERTWO",
+}
 
+export const calculateWinner = (
```
vithushan19, line 15: Right now this function returns void. To make it more testable it should return a string and you can write a test to assert that the return value is what you expect for a given input test case.

If you're still planning to change the logic for this calculation then it's fine to add the unit test later. But if this is the expected logic then you should add the unit test in your PR. That is the best practice that most engineering teams follow. 

At duolingo, there was a rule that almost every PR had to have at least one accompanying test.

In interviews you should say that you do that to make sure your code is always working as expected.

If you add tests now and you made a mistake in your logic then your PR would fail. If you add tests later, you could be making a mistake now and not catch it until much later.

Not a big deal for this project, but in general the longer it takes for a business to identify a bug the more costly that mistake can be.

It's easiest to fix bugs before the code is merged into main vs after you've merged and started thinking about your next task.

I don't think you have a bug in your code, but this defensive mentality is the mindset you need to have for interviews.

pages/studentPortal/labs/multiplication-connect/Index.tsx
```diff
@@ -35,69 +25,121 @@ export const calculateWinner = (grid: GameBoardBlock[]) => {
...
-        rows[i][index].isSelected &&
-        rows[i - 1][index - 1].isSelected &&
-        rows[i - 2][index - 2].isSelected &&
-        rows[i - 3][index - 3].isSelected
-          ? console.log("(descending diagonal) Four in a row!")
+        rows[i][index].selectedBy == player &&
+        rows[i - 1][index - 1].selectedBy == player &&
+        rows[i - 2][index - 2].selectedBy == player &&
+        rows[i - 3][index - 3].selectedBy == player
+          ? console.log(player, "(descending diagonal) Four in a row!")
```
vithushan19, line 71: You should return the string here instead of console logging. Then you can test the output of this function

vithushan19 (approved):
Great! Merge this in and start thinking about how you move the state into a redux slice with actions.

I would talk to Raveen before you start redux

pages/studentPortal/labs/multiplication-connect/Index.tsx
kavanparam: Yeah for sure, still unfinished. Just logging to the console for the time being. I will implement the string return and a test cases in a PR soon.

PR #780, opened 2022-09-24
Gs coding challenges

gajans12345:


vithushan19 (approved):
These look great! Super creative with the pokemon challenges. Very nice! Going forward we're hosting our challenges on codesandbox. Can you use that and just add a link to our website for your challenges?

Something like this
https://codesandbox.io/embed/greetings-forked-n9l8s1?codemirror=1

You can copy these challenges in your PR into new sandboxes

PR #794, opened 2022-10-12
add goals sidebar icon

kwiedenhaupt:
Screen Shot 2022-10-12 at 6 18 09 PM

added appropriate goals sidebar icon making the page more easily accessible

### DIFF
--- a/components/coding/studentPortal/Sidebar.tsx
+++ b/components/coding/studentPortal/Sidebar.tsx
@@ -138,6 +138,26 @@ export const Sidebar: React.FC<SidebarProps> = ({}: SidebarProps) => {
           }
         />
         <SidebarItem
+          name={"Goals"}
+          link={"/goals"}
+          page={"goals"}
+          icon={
+            <svg
+              className="h-5 w-5 mr-4 text-black"
+              viewBox="0 0 24 24"
+              fill="none"
+              stroke="currentColor"
+              stroke-width="2"
+              stroke-linecap="round"
+              stroke-linejoin="round"
+            >
+              {" "}
+              <path d="M22 11.08V12a10 10 0 1 1-5.93-9.14" />{" "}
+              <polyline points="22 4 12 14.01 9 11.01" />
+            </svg>
+          }
+        />
+        <SidebarItem
           name={"Workshops"}
           link={"/workshops"}
           page={"workshops"}


iamnotreddy:
Good stuff, quick work! This is my bad but we should merge this PR last since this change will expose the page and effectively "launch" the feature. I'd open up a separate branch off main to work on the other tickets and come back to this when we're ready to launch.

A couple notes for when you get back to this:
- let's use the solid colored heroicon to be consistent with the rest of the sidebar
- you need to modify the Sidebar redux slice so the active page is set correctly, see below for an example:

Dashboard is active page, highlighted orange
Screen Shot 2022-10-12 at 11 32 05 AM

Goals is active page, not highlighted orange:
Screen Shot 2022-10-12 at 11 32 57 AM

vithushan19 (approved):
You can merge in PRs if @iamnotreddy approves them. He has more context around this feature than me. 

I'll still review when I can.

vithushan19 (changes requested):
I just saw Raveens comment. I'll block this PR from being approved until you and he both feel that the feature is launch ready.

components/coding/studentPortal/Sidebar.tsx
```diff
@@ -1,16 +1,18 @@
...
+import { FETCH_USER_GOALS } from "../../../graphql/fetchUserGoals";
 import { useAuth } from "../../../lib/authContext";
 import {
   activePageSelector,
   setActivePage,
   SidebarProps,
 } from "../../../redux/sidebarSlice";
 
-const SidebarItem = ({ name, link, page, icon }) => {
+const SidebarItem = ({ name, link, page, icon, number }) => {
```
iamnotreddy, line 23: I don't think this is a good approach because we only want to incorporate a number of "urgent" goals for just the Goals sidebar link, not every other page

let's take a step back and brainstorm how to approach this (i honestly don't know the answer to that right now)... also let's get the CEO's design approval before we make a big UI change 😛 

@vithushan19 - for context, we're trying to display some sort of visual cue if a user has X goals due within the next week.... maybe something like this: 
Screen Shot 2022-10-16 at 8 56 58 PM

components/coding/studentPortal/Sidebar.tsx
```diff
@@ -51,6 +54,19 @@ export const Sidebar: React.FC<SidebarProps> = ({}: SidebarProps) => {
     }
   }, [router.pathname]);
 
+  const { data } = useQuery(FETCH_USER_GOALS, {
```
iamnotreddy: i think a better approach would be to do the aggregation on the GQL side... i.e create a new query like FETCH_UPCOMING_GOALS that sums the number of goals due within the following week for that user

components/coding/studentPortal/Sidebar.tsx
```diff
@@ -51,6 +54,19 @@ export const Sidebar: React.FC<SidebarProps> = ({}: SidebarProps) => {
...
+    variables: {
+      userId: user.uid,
+    },
+  });
+
+  const [number, setNumber] = useState();
+  useEffect(() => {
```
iamnotreddy: this state var and useEffect won't be necessary if we get the aggregated number of upcoming goals from GQL

iamnotreddy:
@kwiedenhaupt - was going to leave some comments but let's try to connect tomorrow and walk through these changes, i think there a few different things we need to rework

PR #795, opened 2022-10-12
goal page enhancements

kwiedenhaupt:
Screen Shot 2022-10-12 at 9 45 28 PM

- Expanded columns in grid of md screen width by 2
- Added column "Days Remaining" to be calculated using differenceInCalendarDays date-fn (targetDate-current date)
- set text to red if the goal is within 7 days
- set background to yellow and text to red if goal is within 1 day
Screen Shot 2022-10-12 at 9 39 05 PM

- addressed mobile responsiveness by allowing this new days remaining column to be hidden in mobile view, while still having the highlights operational.

### DIFF
--- a/components/coding/GoalsSectionComponent.tsx
+++ b/components/coding/GoalsSectionComponent.tsx
@@ -37,7 +37,7 @@ export default function GoalsSection({
           <div
             className={`grid grid-cols-5 my-2 text-sm text-center md:grid-cols-12 md:text-lg place-items-center ${
               differenceInCalendarDays(new Date(goal.targetDate), new Date()) <=
-              2
+              1
                 ? "text-red-500 bg-yellow-100"
                 : differenceInCalendarDays(
                     new Date(goal.targetDate),
--- a/components/coding/GoalsSectionComponent.tsx
+++ b/components/coding/GoalsSectionComponent.tsx
@@ -1,11 +1,5 @@
 import { PencilAltIcon, PencilIcon } from "@heroicons/react/outline";
-import {
-  differenceInCalendarDays,
-  differenceInDays,
-  format,
-  formatDistance,
-  isSameWeek,
-} from "date-fns";
+import { differenceInCalendarDays, format } from "date-fns";
 import moment from "moment";
 import Link from "next/link";
 import React from "react";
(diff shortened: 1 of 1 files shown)


vithushan19 (commented):
Looks good to me, but I'll let Raveen give the approval in case he has comments

components/coding/GoalsSectionComponent.tsx
```diff
@@ -16,17 +23,30 @@ export default function GoalsSection({
...
           <p className="hidden md:block md:col-span-2">Date Added</p>
           <p className="font-semibold md:col-span-2">Target Date</p>
+          <p className="hidden md:block md:col-span-2">Days Remaining</p>
         </div>
       )}
 
       {userGoals.map((goal, index) => {
         return (
-          <div className="grid grid-cols-5 my-2 text-sm text-center md:grid-cols-10 md:text-lg place-items-center">
+          <div
```
iamnotreddy, line 37: logic is perfect but I think we can clean this up by defining the logic outside the component so our JSX is cleaner... maybe something like goalStyle = {differenceInCalendarDays(...)}, then reference goalStyle inside className

components/coding/GoalsSectionComponent.tsx
```diff
@@ -35,7 +55,9 @@ export default function GoalsSection({
             <p className="md:col-span-2">
               {format(new Date(goal.targetDate), "MM/dd/yyyy")}
             </p>
-
+            <p className="hidden md:block md:col-span-2">
```
iamnotreddy, line 58: love the consideration for mobile responsiveness :)

iamnotreddy:
Great work! I'd say address that one comment on moving the logic outside of the functional component and we're good to merge this

iamnotreddy (changes requested):
let's clean up this one small piece of logic and we're ready to go -

logic is perfect but I think we can clean this up by defining the logic outside the component so our JSX is cleaner... maybe something like goalStyle = {differenceInCalendarDays(...)}, then reference goalStyle inside className

components/coding/GoalsSectionComponent.tsx
kwiedenhaupt: did that. thank you!

kwiedenhaupt:
- added differenceInDays variable to keep from repeating code.
- added goalStyles variable to keep tailwind styling cleaner

iamnotreddy (approved):

PR #797, opened 2022-10-19
react props tutorial

iamnotreddy:
added a tutorial for props that builds off the previous components tutorial, here's the high level lesson flow: 
- introduce the prop concept in a very simple Hello World component
- bring back Expense Table example from previous lesson and create a "ExpenseRow" component with props
- implement ExpenseRow components in a CodeSandbox
- end lesson by moving expense rows into array of objects, and mapping through that array to return a set of ExpenseRow component

FYI - i'm keeping these new lessons in the warmups folder until we expose / launch them

A couple screencaps: 

Screen Shot 2022-10-19 at 12 12 48 PM
a58-4f50-875f-fd51da61a9c1.png">


vithushan19 (commented):
wow this is actually really good. I think this is exactly what our new students will need as they jump from JS to React

vithushan19 (approved):

PR #800, opened 2022-10-21
visual enhancements to ProfileGoalsSection component

kwiedenhaupt:
1.  added padding to the top
2.  sliced the array so that only the first five goals map on the screen
Currently it displays the first five goals.  It would be possible to do the most recent, but it would make sense that the oldest ones should be addressed first I would think?

### DIFF
--- a/components/coding/ProfileGoalsSection.tsx
+++ b/components/coding/ProfileGoalsSection.tsx
@@ -24,7 +24,10 @@ export default function ProfileGoalsSection({ user }: GoalsSectionProps) {
       },
     }
   );
-  const slicedUserGoals = userGoals.slice(0, 5);
+  const currentUserGoals = userGoals.filter(
+    (goal) => !goal.isComplete && !goal.isArchived
+  );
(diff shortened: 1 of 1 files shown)


kwiedenhaupt:
Screen Shot 2022-10-21 at 9 47 12 PM

kwiedenhaupt:
Screen Shot 2022-10-21 at 9 58 27 PM

This is the total number of goals I have: where only five are showing on the profile page.

components/coding/ProfileGoalsSection.tsx
```diff
@@ -24,6 +24,7 @@ export default function ProfileGoalsSection({ user }: GoalsSectionProps) {
       },
     }
   );
+  const slicedUserGoals = userGoals.slice(0, 5);
```
iamnotreddy, line 27: one small change I'm thinking here - how about we filter to only show current goals, but keep the slice so we show 5 current goals at most? 

I think this makes more sense from a user perspective - what do you think?

iamnotreddy (approved):
Check my one comment but I'm approving

PR #802, opened 2022-10-25
launch sidebar

kwiedenhaupt:
Screen Shot 2022-10-25 at 8 11 54 PM

1. created userGoals icon and added it to the SideBar and sideBarSlice
2. created notification dot to alert upcoming goals due within 7 days
3. implemented query -- fetchQueryUserGoalsCount in Layout.tsx and called it in the Sidebar
4. added isGoalsApproaching to userGoalsSlice as a boolean to save whether or not a notification should be displayed
5. added Sidebar props in order to make the notifications prop a conditional render
6. please note: notifications does still have a red line under it due to the fact it says the payload action is string and the prop requires a boolean.  The app runs -- and the slice defines isGoalsApproaching as a boolean payload. Does it have to do with Dispatch type shows a payload of string? and that is standard?


components/coding/studentPortal/Layout.tsx
```diff
@@ -1,10 +1,37 @@
...
+  const goalApproaching = useSelector(userGoalsSelector);
+  const dispatch = useDispatch();
+  const { data } = useQuery<FetchGoalCountResponse>(FETCH_USER_GOALS_COUNT, {
+    variables: {
+      userId: user.uid,
+      goalDateThreshold: goalDateThreshold,
+    },
+    onCompleted: (data: FetchGoalCountResponse) => {
+      if (data.user_goals_aggregate != null) {
+        dispatch(isGoalApproaching(goalApproaching));
```
vithushan19, line 31: You're not using the data coming back from the query? You're just dispatching an action with the existing value in the selector?

components/coding/studentPortal/Sidebar.tsx
```diff
@@ -150,6 +164,26 @@ export const Sidebar: React.FC<SidebarProps> = ({}: SidebarProps) => {
             </svg>
           }
         />
+        <SidebarItem
+          name={"Goals"}
+          notifications={dispatch(isGoalApproaching(goalApproaching))}
```
iamnotreddy, line 169: you shouldn't be dispatching an action here - you want to pass the Sidebar component a boolean value on whether a goal is approaching

components/coding/studentPortal/Layout.tsx
iamnotreddy: I think the kind of logic you're looking for here is:
- does the FETCH_USER_GOALS_COUNT query return a response i.e is there any upcoming goals? (you're doing this)
- if so, you want to dispatch an action to set goalApproaching to true

the way i'm reading your logic right now is: set the goalApproaching value to the value of goalApproaching (i.e never change it)

redux/userGoalsSlice.ts
```diff
@@ -31,6 +33,12 @@ export const userGoalsSlice: Slice = createSlice({
       }
     },
 
+    isGoalApproaching: (state, action: PayloadAction<boolean>) => {
```
iamnotreddy, line 36: i think we should change the name here so it's more obvious that this reducer is a function that sets a value i.e setIsGoalApproaching

when I read isGoalApproaching, i think this is a boolean variable

iamnotreddy (commented):
@kwiedenhaupt - take a look at mine / vithushan's comments, I pulled this locally and it's not fully working yet

- you need to fix how you're dispatching the action to isGoalApproaching from the query result, and fix how you're calling the selector in the SideBar component; also i'd change the naming of the reducer (setter function) because it's easy to confuse it with the selector

- this is somewhat of a nitpick, but I think it makes more sense organizationally to include isGoalApproaching and its corresponding reducer in the Sidebar slice since the notifications dot is part of the Sidebar feature... you could do it either way though so not required 

- The dot looks great! One small thing that we can maybe address in a different PR if it's not an obvious fix: when you a hover over the Goals tab and it highlights orange / shifts, the dot doesn't move which makes the spacing a bit weird looking

components/coding/studentPortal/Layout.tsx
kwiedenhaupt: fixed.

components/coding/studentPortal/Sidebar.tsx
kwiedenhaupt: changed to only goalApproaching, which is the boolean value

redux/userGoalsSlice.ts
kwiedenhaupt: name is changed to setIsGoalApproaching

iamnotreddy (approved):
changes look good!

PR #803, opened 2022-10-28
add box for notes on a goal when editing

kwiedenhaupt:
Screen Shot 2022-10-28 at 8 03 38 PM

1. added box for the notes
2. edited corresponding queries, upsert, and type
3. added corresponding graphQL column in user_goals table to save data

### DIFF
--- a/pages/goals/[slug].tsx
+++ b/pages/goals/[slug].tsx
@@ -76,6 +76,19 @@ const EditGoalsPage = () => {
                 }));
               }}
             />
+            <p className="font-bold">Write An Actionable Goal Outline</p>
+
+            <textarea
+              className={`text-left p-2 border rounded-md shadow-md w-full md:w-1/2 text-murkrow `}
+              placeholder={goalDetail.goalNotes}
+              value={editedGoalValues.goalNotes}
+              onChange={(e) => {
+                setEditedGoalValues((prevState) => ({
+                  ...prevState,
+                  goalName: e.target.value,
+                }));
+              }}
+            />
             {editedGoalValues.goalName.length > 60 && (
               <p className="text-xs text-red-600">
                 please keep your goal under 60 characters
--- a/pages/goals/[slug].tsx
+++ b/pages/goals/[slug].tsx
@@ -76,6 +76,12 @@ const EditGoalsPage = () => {
                 }));
               }}
             />
+
+            {editedGoalValues.goalName.length > 60 && (
+              <p className="text-xs text-red-600">
+                please keep your goal under 60 characters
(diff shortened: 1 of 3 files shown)


iamnotreddy (commented):
Need to add goalNotes to the editedGoalValuesForHasura - for context, see the note for why this field is necessary

There's something else going on though, wasn't able to get to mutation to write back successfully. Lmk if you can figure it out and get a note to write back successfully to Hasura - I'll try to take a look later

`                  const editedGoalValuesForHasura = {
                    goalName: editedGoalValues.goalName,
                    userId: editedGoalValues.userId,
                    id: editedGoalValues.id,
                    isArchived: editedGoalValues.isArchived,
                    isComplete: editedGoalValues.isComplete,
                    targetDate: editedGoalValues.targetDate,
                  };`

iamnotreddy:
@kwiedenhaupt  - pretty sure I figured out the issue here, it was not super obvious. Gonna share how I went about solving this issue. cc @vithushan19 - something to be aware of when adding new fields in hasura tables.

The issue I kept getting in VSCode when trying to save a new goal note implied that the mutation `user_goals_update_column` wasn't recognizing the goalNotes field which we recently added: 
image

I reproduced the issue inside the Hasura console it was saying the same thing: 
image

So the mutation isn't recognizing our new field even though we updated all the queries, types, apollo functions etc.. Did some googling and found people running into a similar [issue](https://github.com/apollographql/apollo-client/issues/5080). The comments suggested I try adding / deleting a permission on this table. This apparently resets the graphQL schema and the update mutation now recognizes are new field. No idea why this works lol, but good to note for the future.

I was able to add a note with no errors. Can you confirm this on your side as well?

pages/goals/[slug].tsx
```diff
@@ -76,11 +76,26 @@ const EditGoalsPage = () => {
                 }));
               }}
             />
+
             {editedGoalValues.goalName.length > 60 && (
               <p className="text-xs text-red-600">
                 please keep your goal under 60 characters
               </p>
             )}
+            <p className="font-bold">Write An Actionable Goal Outline</p>
```
Comment (iamnotreddy) [inline pages/goals/[slug].tsx:89]: purely editorial suggestion here - what if we stick to "goal notes"? it seems less prescriptive i.e write whatever you want here associated with the goal.

what if we put "e.g write an actionable goal outline" in the placeholder if notes are empty instead?

Comment (kwiedenhaupt) [reply pages/goals/[slug].tsx]: that's even better. just made the change

iamnotreddy (approved):
approved - nice work on this!

PR #805, opened 2022-10-30
add Clear text button to Goal Notes

kwiedenhaupt:
Not sure if this feature is necessary.  Used the useRef hook and set it to null so that the notes box can be cleared by the user if desired. 
Also set the local state variable to match that of the icon that disables the input box so that the button is disabled as well.
Screen Shot 2022-10-30 at 9 26 35 PM
Screen Shot 2022-10-30 at 9 26 18 PM


LuckyHariharan:
Clean and clear pull request and code. The feature looks helpful to me as a user. Would definitely prefer the option to clear a lengthy Actionable Goal Outline as opposed to clearing by hand. LGTM!

PR #806, opened 2022-11-01
useState lesson

iamnotreddy:
@vithushan19 - got most of the lesson on paper, but still need to clean up / add a few more things.. lmk what you think / feel free to add anything to this

will finish this up when i'm back


vithushan19 (approved):
Looks good to me. Dont have anything to add personally. We can show it to new students and get feedback

PR #810, opened 2022-11-09
added loom video component

kwiedenhaupt:
Screen Shot 2022-11-09 at 9 06 32 PM
Added code for the component on both the template file and the AssignmentComponent -- the key is to paste in the template file only the end of the embedded link identifier for the video.


components/coding/studentPortal/AssignmentComponent.tsx
```diff
@@ -2,6 +2,7 @@ import React, { useState } from "react";
 import { Sandpack } from "@codesandbox/sandpack-react";
 import HintRow from "./HintRow";
 import AssignmentInputBox from "./AssignmentInputBox";
+import { Stringifier } from "postcss";
```
iamnotreddy, line 5: don't need this

iamnotreddy (commented):
- remove the fetchUserAssignmentSubmissions file
- remove the Stringifier import

components/coding/studentPortal/AssignmentComponent.tsx
kwiedenhaupt: removed.

iamnotreddy (approved):
looks good! i was trying to see if we can clean up the commit log history since those messages are already in main but i think that will be too tedious. 

approved but can you remove that Stringifier import in the AssignmentComponent file

PR #811, opened 2022-11-11
added stages to AssignmentComponent

kwiedenhaupt:
Amended template.tsx file to add stages: 
1. Incomplete
2. Submitted
3. Completed

Using the Previous/Next buttons to cycle through the stages using the useState hook to remember them.

Added a clear functionality and an alert to the save button

added the proper data for the components
Screen Shot 2022-11-12 at 12 06 10 AM
Screen Shot 2022-11-12 at 12 05 34 AM
Screen Shot 2022-11-12 at 12 05 45 AM


components/coding/studentPortal/AssignmentInputBox.tsx
```diff
@@ -49,7 +50,13 @@ export const AssignmentInputBox: React.FC<AssignmentInputBoxProps> = ({
               variables: {
                 objects: submissionVariables,
               },
+              onCompleted: () => {
```
iamnotreddy, line 53: Not incorrect but would be cleaner if you move this onComplete above where you call useMutation to define the saveAssignmentInput function. You can also move the variables object up there

pages/studentPortal/web/React/assignments/template.tsx
```diff
@@ -1,33 +1,60 @@
...
   };
+  const [stage, setStage] = useState(0);
+
   return (
     <>
       <div className="grid grid-cols-1 gap-8 px-4 pt-4 m-8 sm:px-12">
-        {assignmentComponents.map((it: AssignmentComponentData) => (
-          <AssignmentComponent data={it} />
-        ))}
+        {stage === Stage.INCOMPLETE
```
iamnotreddy, line 32: 🔥🔥🔥

iamnotreddy (approved):
Still going to see whether we need the Promise construction but otherwise approved since everything is working. I'll comment on that when I get a chance to check.

Really awesome work on incorporating the Stages pattern within the lesson component framework that we've been re-using for assignments.

Take note of your commit history on this PR - it includes all the commits from the last two PRs. This shouldn't be happening. Let's spend some more time next week checking your process with git

components/coding/studentPortal/AssignmentInputBox.tsx
kwiedenhaupt: Done

PR #812, opened 2022-11-13
Added useQuery to send student to appropriate assignment stage

kwiedenhaupt:
1. created fetchUserAssignmentSubmissions query that fetches data from user_assignment_submissions table
2. called the query in template.tsx and filtered the data for the submission_link field
3. logic in useEffect hook to run at the end then sends user to the appropriate stage based on whether submission_link contains data.

### DIFF
--- a/graphql/fetchUserAssignmentSubmissions.ts
+++ b/graphql/fetchUserAssignmentSubmissions.ts
@@ -1,0 +1,20 @@
+import { gql } from "@apollo/client";
+export const FETCH_USER_ASSIGNMENT_SUBMISSIONS = gql`
+  query fetchUserAssignmentSubmissions($user_id: String = "") {
+    user_assignment_submissions(where: { user_id: { _eq: $user_id } }) {
+      submission_link
+      id
+      user_id
+      last_updated
+    }
+  }
+`;
+export type FetchUserAssignmentSubmissionsDataResponse = {
+  user_assignment_submissions: Array<UserAssignmentSubmissions>;
+};
+export type UserAssignmentSubmissions = {
+  id: string;
+  user_id: string;
+  submission_link: string;
+  last_updated: Date;
+};
--- a/graphql/fetchUserAssignmentSubmissions.ts
+++ b/graphql/fetchUserAssignmentSubmissions.ts
@@ -6,6 +6,7 @@ export const FETCH_USER_ASSIGNMENT_SUBMISSIONS = gql`
       id
       user_id
       last_updated
+      review_link
     }
   }
 `;
@@ -17,4 +18,5 @@ export type UserAssignmentSubmissionsData = {
   user_id: string;
   submission_link: string;
   last_updated: Date;
+  review_link: string;
 };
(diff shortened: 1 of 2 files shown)


pages/studentPortal/web/React/assignments/template.tsx
```diff
@@ -39,7 +39,7 @@ const React2 = ({ incompleteStage, submittedStage, completedStage }) => {
     let submittedAssignment = assignments.filter(
       (assignment) => assignment.submission_link
     );
-    if (submittedAssignment) {
+    if (submittedAssignment.length > 0) {
```
kwiedenhaupt, line 42: This query does work...and if I put in the following div, returns the information that I want: 
 ```
<div>
        {assignments.map((assignment) => (
          <div>{assignment.submission_link}</div>
        ))}
      </div>
```
I am unclear, however, why this query returns [object, object] and the field __typename and therefore does not allow me to print out data or data.user_assignment_submissions etc.

wondering if there is a better way?

File: graphql/fetchUserAssignmentSubmissions.ts [hunk 2]
```diff
@@ -0,0 +1,20 @@
+import { gql } from "@apollo/client";
+export const FETCH_USER_ASSIGNMENT_SUBMISSIONS = gql`
+  query fetchUserAssignmentSubmissions($user_id: String = "") {
+    user_assignment_submissions(where: { user_id: { _eq: $user_id } }) {
+      submission_link
+      id
```
kwiedenhaupt, line 6: I pulled data for all the fields in the table in graphQL. wasn't sure if that was good so that I could use the same query later if needed or if it would be better just to query and set types for the one field I needed for this task.

pages/studentPortal/web/React/assignments/template.tsx
```diff
@@ -1,15 +1,48 @@
...
+
+      onCompleted: (data: FetchUserAssignmentSubmissionsDataResponse) => {
+        setAssignments(data.user_assignment_submissions);
+      },
+    }
+  );
+  useEffect(() => {
+    if (assignments) {
+      setStage(Stage.SUBMITTED);
+    } else {
```
kwiedenhaupt, line 44: Is it better to use the useEffect here or just to put in the onCompleted as well?

graphql/fetchUserAssignmentSubmissions.ts
iamnotreddy: see comment below about the response being an array

iamnotreddy (changes requested):

pages/studentPortal/web/React/assignments/template.tsx
```diff
@@ -1,15 +1,46 @@
...
+  const { data } = useQuery<FetchUserAssignmentSubmissionsDataResponse>(
+    FETCH_USER_ASSIGNMENT_SUBMISSIONS,
+    {
+      variables: {
+        user_id: user.uid,
+      },
+
+      onCompleted: (data: FetchUserAssignmentSubmissionsDataResponse) => {
+        setAssignments(data.user_assignment_submissions);
+        if (data.user_assignment_submissions[0].submission_link.length > 0) {
```
iamnotreddy, line 35: lets add review link to this logic

kwiedenhaupt: done.

iamnotreddy (approved):
looks good now - awesome work on this!

PR #813, opened 2022-11-15
notifications for web development assignments

kwiedenhaupt:
WIP regarding the feature, because the graphQL table needs to be updated through queries rather than manually, as it is now, but this PR creates a notification dot on the Sidebar page determined by that boolean value.

Screen Shot 2022-11-15 at 8 46 49 PM

### DIFF
--- a/components/coding/studentPortal/Layout.tsx
+++ b/components/coding/studentPortal/Layout.tsx
@@ -3,12 +3,17 @@ import { format, addDays } from "date-fns";
 import React, { useState } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import {
+  FetchUserAssignmentSubmissionsDataResponse,
+  FETCH_USER_ASSIGNMENT_SUBMISSIONS,
+} from "../../../graphql/fetchUserAssignmentSubmissions";
+import {
   FetchGoalCountResponse,
   FETCH_USER_GOALS_COUNT,
 } from "../../../graphql/fetchUserGoalsCount";
 import { useAuth } from "../../../lib/authContext";
 import {
   activePageSelector,
+  setAssignmentReviewed,
   setIsGoalApproaching,
 } from "../../../redux/sidebarSlice";
 import Navbar from "../../ui/Navbar";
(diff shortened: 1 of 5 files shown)


components/coding/studentPortal/Layout.tsx
```diff
@@ -36,6 +41,28 @@ export const Layout: React.FC = ({ children }) => {
...
 
+  const { data } = useQuery<FetchUserAssignmentSubmissionsDataResponse>(
+    FETCH_USER_ASSIGNMENT_SUBMISSIONS,
+    {
+      variables: {
+        user_id: user.uid,
+      },
+      onCompleted: (data) => {
+        if (data.user_assignment_submissions) {
+          let instructorReviewed = data.user_assignment_submissions.filter(
```
iamnotreddy, line 52: you can declare this variable with const since you're not re-assigning the variable anywhere else

components/coding/studentPortal/Sidebar.tsx
```diff
@@ -72,6 +73,16 @@ export const Sidebar: React.FC<SidebarProps> = ({}: SidebarProps) => {
     }
   }, [router.pathname]);
 
+  const returnWebDevStyle = (assignmentReviewed: boolean) => {
```
iamnotreddy, line 76: nice!

iamnotreddy (approved):
looks good, approved!

one piece of feedback if you have time - I think we can move the returnWebDevStyle function to a separate file and import it in (maybe create api/studentPortal/helpers.ts and store it there), and potentially move the logic inside the useQuery onCompleted into a helper function as well. 

layout and sidebar are pretty important skillify files, so think it's good to keep them as light / clean as possible

components/coding/studentPortal/Layout.tsx
```diff
@@ -36,6 +41,28 @@ export const Layout: React.FC = ({ children }) => {
     fetchPolicy: "cache-and-network",
   });
 
+  const { data } = useQuery<FetchUserAssignmentSubmissionsDataResponse>(
+    FETCH_USER_ASSIGNMENT_SUBMISSIONS,
+    {
+      variables: {
+        user_id: user.uid,
+      },
```
iamnotreddy, line 49: add assignmentId here when I'm finished

PR #814, opened 2022-11-17
modify assignments workflow to prevent more than one submission per user

iamnotreddy:
This PR contains a few changes to account for assignmentId being added to Hasura:

- the coding_assignments table is now part of the whole assignments workflow - this table will have a record of every assignment
- there is now an assignment_id field in the user_assignment_submissions tables - it is foreign key to coding_assignments... we will only allow submissions with a valid assignment_id
- from a coach or content creator perspective, they must create a record of the assignment inside coding_assignments which generates an assignment_id uuid 
- for now, we store this key when creating a new assignment, then pass it into both the fetch / mutation queries when pulling or updating a student's assignment submissions 
- this solves for the following: 1) we now only check whether a submission has been made for that **specific** assignment (before it was all assignments) when determining which stage to render and 2) a user can only make one submission per assignment - if they change their submission, it only updates the submission link

### DIFF
--- a/components/coding/studentPortal/AssignmentComponent.tsx
+++ b/components/coding/studentPortal/AssignmentComponent.tsx
@@ -37,6 +37,7 @@ export type AssignmentComponentData =
       codeSandboxTitle: string;
       link: string;
       placeholder: string;
+      assignmentId: string;
     }
   | {
       component: "custom";
@@ -51,7 +52,6 @@ export type AssignmentComponentData =
       component: "loom-video";
       text?: string;
       videoId: string;
-
     }
   | {
       component: "completed";
@@ -97,7 +97,8 @@ export default function AssignmentComponent({
         <h1 className="font-bold">{data.codeSandboxTitle}</h1>
         <AssignmentInputBox
           placeholder={data.placeholder}
-          submission_link={""}
+          submissionLink={""}
+          assignmentId={data.assignmentId}
         />{" "}
       </>
     );
@@ -119,7 +120,7 @@ export default function AssignmentComponent({
           moz-allowfullscreen
           allowFullScreen
           className="w-full h-96"
-        ></iframe>
+        />
       </div>
     );
   } else {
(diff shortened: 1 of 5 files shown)


components/coding/studentPortal/AssignmentComponent.tsx
```diff
@@ -97,7 +97,8 @@ export default function AssignmentComponent({
         <h1 className="font-bold">{data.codeSandboxTitle}</h1>
         <AssignmentInputBox
           placeholder={data.placeholder}
-          submission_link={""}
+          submissionLink={""}
+          assignmentId={data.assignmentId}
```
kwiedenhaupt, line 101: This is cool how you passed it down as a prop here.

components/coding/studentPortal/AssignmentInputBox.tsx
```diff
@@ -1,24 +1,27 @@
...
 import React, { useState } from "react";
+import { FETCH_USER_ASSIGNMENT_SUBMISSIONS } from "../../../graphql/fetchUserAssignmentSubmissions";
 import { UPSERT_USER_ASSIGNMENT_SUBMISSIONS } from "../../../graphql/upsertUserAssignmentSubmissions";
 import { useAuth } from "../../../lib/authContext";
 import { Button } from "../../ui/Button";
 
 export interface AssignmentInputBoxProps {
   placeholder: string;
-  submission_link: string;
+  submissionLink: string;
```
kwiedenhaupt, line 11: cleaner without the underscore I think.

graphql/upsertUserAssignmentSubmissions.ts
```diff
@@ -10,7 +10,7 @@ export const UPSERT_USER_ASSIGNMENT_SUBMISSIONS = gql`
     insert_user_assignment_submissions(
       objects: $objects
       on_conflict: {
-        constraint: user_assignment_submissions_pkey
+        constraint: user_assignment_submissions_user_id_assignment_id_key
```
kwiedenhaupt, line 13: Amazing. This will block multiple submissions in the table for the same person/assignment.

kwiedenhaupt (approved):
Very well thought out for this stage in developing the feature. You're teaching me a lot reading through your work!

PR #815, opened 2022-11-17
add recipes page

iamnotreddy:
created a recipe page reusing the resource-list component from the lesson component library.. only two recipes on here so far

lmk where you think this should live and I can surface it in another commit

### DIFF
--- a/pages/recipes.tsx
+++ b/pages/recipes.tsx
@@ -1,0 +1,46 @@
+import React from "react";
+import LessonComponent, {
+  LessonComponentData,
+} from "../components/coding/studentPortal/LessonComponent";
+
+const React2 = ({ lessonComponents }) => {
+  return (
+    <>
+      <div className="grid grid-cols-1 gap-8 px-4 pt-4 m-8 sm:px-12">
+        {lessonComponents.map((it) => (
+          <LessonComponent data={it} />
+        ))}
+      </div>
+    </>
+  );
+};
+
+export async function getServerSideProps({ params }) {
+  const lessonComponents: LessonComponentData[] = [
+    {
+      component: "title",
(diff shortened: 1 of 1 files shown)


vithushan19:
Looks good to me, I might slot this in the react lessons under lesson 1, components. 
https://github.com/skillify-ca/skillify-web/tree/main/pages/studentPortal/web/React

Maybe we should create a components folder in the react folder. Right now if you click to continue from Lesson 1 components, it takes you to lesson 2 Props. But instead, we should take the user to this recipes page. 

I don't think there should be a separate page where we keep all the recipes. It's another destination that students will be confused by. It's better if it's just part of the linear student experience of going through react lessons. 

I'm thinking there might be value in being able to search for a recipe, but we could achieve that with a search box in the lesson timeline. 

Even with the workshops, I'm thinking about how to integrate them better into the JavaScript lessons so it's a linear path, rather than another destination. Students get overwhelmed when they see so many different destinations. 

That was the reason I took the math code out of the labs page. I want to remove the labs page from the sidebar completely. 

I'm also thinking about how we can structure our content better using Sanity: https://www.sanity.io/structured-content

let me know if you want to discuss further, but i'm fine to add these recipes into the existing components lesson, (atleast for your first recipe) 

the second recipe could be added to a new lesson about input forms: https://beta.reactjs.org/learn/manipulating-the-dom-with-refs#example-focusing-a-text-input

vithushan19 (approved):

iamnotreddy:
I'm onboard with putting it on the components instead of lessons page - all your reasoning makes sense, did feel like the Sidebar was already getting crowded.

And I totally agree with finding a better structure - current process doesn't feel super sustainable if we generate a lot more content

PR #816, opened 2022-11-30
small changes to UI of assignment component

kwiedenhaupt:
Screen Shot 2022-11-30 at 11 23 29 AM

These changes included removing unneeded spacing and moving buttons to the left for easier use/nicer look.


iamnotreddy (approved):
nice - this looks much cleaner! 

great work on this feature - we're definitely at a point where I can create an actual assignment... will look to do that in the next week or two, at which point I'll be able to give more feedback on any other changes

PR #818, opened 2022-12-02
v2 profile page and ProfileHeaderComponent

kwiedenhaupt:
1. created v2 page in order to visualize work
2. created ProfileHeaderComponent to look like Figma outline
3. created query to display total badges possible to fit Figma sketch
4. worked off the mobile view and then added the tailwind md styling to make things bigger for the regular computer screen

Screen Shot 2022-12-02 at 3 31 38 PM
Screen Shot 2022-12-02 at 3 33 01 PM

### DIFF
--- a/components/coding/ProfileHeaderComponent.tsx
+++ b/components/coding/ProfileHeaderComponent.tsx
@@ -1,0 +1,113 @@
+import { useQuery } from "@apollo/client";
+import { format } from "date-fns";
+import React, { useState } from "react";
+import {
+  FetchUserBadgesCountResponse,
+  FETCH_USER_BADGES_COUNT,
+} from "../../graphql/fetchUserBadgesCount";
+
+import {
+  FetchUserProfileDataResponse,
+  FETCH_USER_PROFILE_DATA,
+  UserProfileData,
+} from "../../graphql/fetchUserProfile";
+
+export type UserProfileSectionProps = {
+  user: any;
+};
+
+export default function UserProfileSection({ user }: UserProfileSectionProps) {
+  const [userProfileData, setUserProfileData] =
+    useState<UserProfileData>(Object);
+
+  const [userBadgeCount, setUserBadgeCount] = useState<number>(0);
+
+  const { loading: userProfileLoading } =
+    useQuery<FetchUserProfileDataResponse>(FETCH_USER_PROFILE_DATA, {
+      variables: {
+        userId: user.uid,
+      },
+      onCompleted: (data) => {
+        if (data.users.length > 0) {
+          setUserProfileData({
+            typeName: data.users[0].__typename,
+            createdAt: data.users[0].created_at,
+            email: data.users[0].email,
+            lastSeen: data.users[0].last_seen,
+            name: data.users[0].name,
+            profileImage: data.users[0].profile_image,
+          });
+        }
+      },
+    });
+  const { loading: userBadgeCountLoading } =
+    useQuery<FetchUserBadgesCountResponse>(FETCH_USER_BADGES_COUNT, {
+      variables: {
+        userId: user.uid,
+      },
+      onCompleted: (data) => {
+        if (data.user_coding_badges_aggregate.aggregate.count) {
+          setUserBadgeCount(data.user_coding_badges_aggregate.aggregate.count);
+        }
+      },
+    });
+
+  return (
+    <>
+      <div className="flex items-center justify-between">
+        {userProfileData && (
+          <h1 className="text-3xl font-bold">{userProfileData.name}</h1>
+        )}
+      </div>
+      {userProfileLoading ? (
+        <div>Loading...</div>
+      ) : (
+        <div className="grid grid-cols-1 mt-12 sm:grid-cols-8">
+          <img
+            className="w-32 rounded-full"
+            src={userProfileData.profileImage}
+          />
+          <div className="col-span-2 mt-2 ml-2">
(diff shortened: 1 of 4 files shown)


pages/profile/v2/[slug].tsx
```diff
@@ -0,0 +1,85 @@
...
+    cache: new InMemoryCache(),
+  });
+  const { data } = await client.query({
+    query: FETCH_USER,
+    variables: {
+      link: params.slug,
+    },
+  });
+
+  console.log(params.slug, data.users[0].id);
```
Comment (iamnotreddy) [inline pages/profile/v2/[slug].tsx:56]: remove this console.log

components/coding/ProfileHeaderComponent.tsx
```diff
@@ -0,0 +1,125 @@
...
+          <img
+            className="w-36 mt-6 mr-2 rounded-full"
+            src={userProfileData.profileImage}
+          />
+          <div className="col-span-3 mt-2 ml-2">
+            <h1 className="text-2xl font-bold mt-2 md:text-4xl md:mt-6">
+              {userProfileData.name}
+            </h1>
+            <div className="flex mt-4">
+              <svg
```
iamnotreddy, line 82: I think we can clean up this component a bit by moving these heroicons into separate components; 

couple ideas / suggestions: 
- perhaps the Joined Date + Icon can be a component that takes in createdAt as props... I've seen this used elsewhere across Skillify
- perhaps the Badges earned out of Total Badges + the Icon can be a component that takes in earned badges and total badges as props... same point as above, think we can reuse this across Skillify

components/coding/ProfileHeaderComponent.tsx
```diff
@@ -0,0 +1,125 @@
...
+} from "../../graphql/fetchUserBadgesCount";
+
+import {
+  FetchUserProfileDataResponse,
+  FETCH_USER_PROFILE_DATA,
+  UserProfileData,
+} from "../../graphql/fetchUserProfile";
+
+export type UserProfileSectionProps = {
+  user: any;
```
iamnotreddy, line 18: let's try to get in the habit of not using the any type unless we have to... I know we're doing it across the codebase but lets try defining a type for this user object so we're using Typescript properly

iamnotreddy (approved):
overall looks good - approved. nice work!

One suggestion before merging - let's think about our file organization structure. There should be a separate folder for profile v2 components i.e the Header. If you split up the joined date+ heroicon and badges + heroicon code  into separate components, you can put them here as well.

Admittedly, we could have done a better job of organization with past work so let's get off to a good start with this project!

components/coding/ProfileHeaderComponent.tsx
kwiedenhaupt: created a User type. great idea.

components/coding/profileV2/BadgesDisplayedComponent.tsx
```diff
@@ -0,0 +1,51 @@
+import { useQuery } from "@apollo/client";
+import { format } from "date-fns";
+import React, { useState } from "react";
+import { FETCH_TOTAL_USER_BADGES_COUNT } from "../../../graphql/fetchTotalUserBadgesCount";
```
iamnotreddy, line 4: let's remove all our unused imports - we aren't using any gql in this component, just passing down props into it so it can render them

vithushan and I were talking about introducing some automated checks to make sure we don't have console.logs or unused imports which would make these things easier to spot

components/coding/profileV2/BadgesDisplayedComponent.tsx
```diff
@@ -0,0 +1,51 @@
...
+import JoinedDateComponent from "./JoinedDateComponent";
+import { User } from "./ProfileHeaderComponent";
+
+export type BadgesDisplayedComponentProps = {
+  user: User;
+  earnedBadges: number;
+  totalBadges: number;
+};
+
+export default function UserProfileSection({
```
iamnotreddy, line 24: Component name should correspond to what this component does

pages/profile/v2/[slug].tsx
```diff
@@ -0,0 +1,85 @@
...
```
Comment (iamnotreddy) [inline pages/profile/v2/[slug].tsx:56]: you forgot to remove the console.log in the new file, but you removed it in the old file (thanks!)

components/coding/profileV2/ProfileHeaderComponent.tsx
```diff
@@ -0,0 +1,124 @@
...
+} from "../../../graphql/fetchUserProfile";
+import BadgesDisplayedComponent from "./BadgesDisplayedComponent";
+
+import JoinedDateComponent from "./JoinedDateComponent";
+
+export type UserProfileSectionProps = {
+  user: User;
+};
+
+export type User = {
```
iamnotreddy, line 22: Nice job mapping this out this type! How about we move this type to the file that contains FETCH_USER_PROFILE_DATA

iamnotreddy (approved):
good work! approved but please make the small tweaks

components/coding/profileV2/BadgesDisplayedComponent.tsx
kwiedenhaupt: that's a great idea! I want to learn how to write those github actions as well:)

components/coding/profileV2/BadgesDisplayedComponent.tsx
kwiedenhaupt: changed to BadgesDisplayedComponent

PR #819, opened 2022-12-04
Basic expandable container component

kwiedenhaupt:
1. created a component using state hooks that expands when clicking the plus/minus in accordance with the product specs.
2. props are "title" and an "open" boolean prop.  
3. added the component to the ProjectsSection.tsx in order to test/visualize my work.
Screen Shot 2022-12-04 at 3 27 28 PM
Screen Shot 2022-12-04 at 3 27 37 PM


components/coding/Expandable.tsx
```diff
@@ -0,0 +1,50 @@
+import React, { useState } from "react";
+
+import { PlusCircleIcon } from "@heroicons/react/solid";
+import { MinusCircleIcon } from "@heroicons/react/solid";
+
+interface ExpandableProps {
+  open?: boolean;
```
iamnotreddy, line 7: I like what you did here - I wasn't originally thinking about supplying open as a prop, but this way we can set whether a component is expanded by default on the parent page

components/coding/Expandable.tsx
```diff
@@ -0,0 +1,50 @@
...
+  title: string;
+}
+
+const ExpandableContainer: React.FC<ExpandableProps> = ({
+  open,
+  children,
+  title,
+}) => {
+  const [isOpen, setIsOpen] = useState(open);
```
iamnotreddy, line 16: The open prop is currently optional. What would happen if `open` is not supplied and we try to set it as the initial value for isOpen? Typescript isn't complaining so this might be ok, but I think it's worth testing that out to make sure.

components/coding/Expandable.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { MinusCircleIcon } from "@heroicons/react/solid";
+
+interface ExpandableProps {
+  open?: boolean;
+  title: string;
+}
+
+const ExpandableContainer: React.FC<ExpandableProps> = ({
+  open,
+  children,
```
iamnotreddy, line 13: I noticed that `children` isn't defined in the expandable props but Typescript knows it's type because it's a property of `React.ReactNode`. 

I probably would have explicitly defined a child property of type React.FC in the ExpandableProps - which I confirmed works but actually seems unnecessary if children is already part of React.FC. So I think what you're doing is right but wanted to point it out to make sure you understand what's going on.

cc @vithushan19 if you have any extra context for us here

components/coding/ProjectsSection.tsx
```diff
@@ -30,7 +31,9 @@ export default function ProjectsSection({ user }: ProjectsSectionProps) {
         <div>Loading...</div>
       ) : userProjects.length === 0 ? (
         <div className="col-span-3 p-8 text-center shadow-md bg-slate-300 dark:bg-slate-900">
-          No Active Projects
+          <ExpandableContainer open title={"No Active Projects"}>
```
iamnotreddy: mentioned this on slack but let's not commit any changes to this component since it's being used on the current profile

iamnotreddy (changes requested):
Nice work putting this together! Before approving, I want to make sure we understand the behavior with that optional open prop.

Also, can we call this file ExpandableContainer? I think that's a little more clear on what the file actually contains.

components/coding/Expandable.tsx
kwiedenhaupt: When open is not supplied it actually defaults to open -- so I think that's what we want it to do more or less.

iamnotreddy: In that case - can we make supplying the open prop mandatory instead of optional? 

If someone was using your component and didn't supply the open prop, it wouldn't have anything to default to. You could add a check to set the initial value to false if it wasn't supplied but that seems more convoluted. 

From a developer standpoint - I think it makes sense to always supply that prop when calling that component

iamnotreddy (approved):
approved - please address that comment on the open prop

great work :)

components/coding/Expandable.tsx
kwiedenhaupt: that's a good idea.  Just made it mandatory

PR #822, opened 2022-12-06
Link to User Profile and Default Profile Image Styling

LuckyHariharan:
Main changes from last PR:

- Updated link field in Hasura, and query to include the link field. Clicking on a users card now directs to skillify.ca/profile/${link}

- Default avatar picture modified to clean up card styling
 
image


components/coding/studentPortal/ProfileDetailCard.tsx
```diff
@@ -10,57 +13,61 @@ type ProfileDetailCard = {
...
-              fill="currentColor"
-              className="w-4 h-4 text-charmander"
-            >
-              <path
-                fill-rule="evenodd"
-                d="M12 2.25c-5.385 0-9.75 4.365-9.75 9.75s4.365 9.75 9.75 9.75 9.75-4.365 9.75-9.75S17.385 2.25 12 2.25zM12.75 6a.75.75 0 00-1.5 0v6c0 .414.336.75.75.75h4.5a.75.75 0 000-1.5h-3.75V6z"
-                clip-rule="evenodd"
-              />
-            </svg>
+      <Link href={profile}>
```
iamnotreddy, line 25: IMO a better pattern would be to wrap the Link component outside the ProfileDetailCard component in the coaches-dashboard parent. I feel like this is more intuitive because if I'm reading the code from that parent page, I know where each component links to without going into that component. I feel like I've seen this pattern more frequently

I'm curious whether this affects the difference in loading time that you noticed after switching from the <a> tag

components/coding/studentPortal/ProfileDetailCard.tsx
```diff
@@ -10,57 +13,61 @@ type ProfileDetailCard = {
...
+        <div>
```
iamnotreddy, line 26: this div is unnecessary if you move the Link component re: comment above

components/coding/studentPortal/ProfileDetailCard.tsx
```diff
@@ -10,57 +13,61 @@ type ProfileDetailCard = {
 function ProfileDetailCard({
   avatar,
   name,
+  link,
   joinDate,
   badges,
   currentFocus,
   nextGoal,
 }: ProfileDetailCard) {
+  let profile = "profile/" + link;
```
iamnotreddy, line 22: this can actually be const because the variable is never reassigned in this component, but it works either way

components/coding/studentPortal/ProfileDetailCard.tsx
```diff
@@ -10,57 +13,61 @@ type ProfileDetailCard = {
...
   return (
     <div className="bg-gray-900 text-sm w-60 p-6 h-full space-y-8 rounded-md ">
```
iamnotreddy, line 24: why is this so dark lol 

is it because you're typically using dark mode?

iamnotreddy (changes requested):
try moving the Link component like I mentioned / clean up that extra div

also - I don't love the hover styling, but I think we're ok with the imperfect styling for now... we will definitely polish this all up before launch

components/coding/studentPortal/ProfileDetailCard.tsx
LuckyHariharan: The cards are dark because I'm having trouble changing the text color from white. Met with Jason about it, he was also having trouble.

components/coding/studentPortal/ProfileDetailCard.tsx
LuckyHariharan: Wrapping the Link tag in the parent component does not seem to speed up the loading time of the link.

components/coding/studentPortal/ProfileDetailCard.tsx
LuckyHariharan: sounds good, deleted the unnecessary div.

iamnotreddy (approved):
nicee work

we'll incorporate Kari's components and work on the styling in the next PR so this looks clean

PR #823, opened 2022-12-07
User Goals Updates

kwiedenhaupt:
Functionality of these changes on the profile/v2 page:
Screen Shot 2022-12-07 at 8 30 39 AM

These changes: 
1. retain the styling for the highlighted goals, etc. that is on the main goals page
2. expand the view of the goal so that it extends to the entire screen
3. Takes in a prop that filters to only current goals and slices to only three goals
4. retains the same functionality for the goals page
5. eliminates the need for ProfileGoalsSection that is used on the main profile page and profile/v2 page -- BUT the profile page would need to be updated like the profile/v2 page with the fetchUserGoals query to populate the data field.

Functionality of these changes on the goals page:
Screen Shot 2022-12-07 at 8 31 59 AM

Eventually the goal would be to implement this GoalsSectionComponent.tsx and get rid of ProfileGoalsSection.tsx.

### DIFF
--- a/components/coding/GoalsSectionComponent.tsx
+++ b/components/coding/GoalsSectionComponent.tsx
@@ -1,17 +1,11 @@
-import { useQuery } from "@apollo/client";
-import { PencilAltIcon, PencilIcon } from "@heroicons/react/outline";
+
+import { PencilAltIcon } from "@heroicons/react/outline";
 import { differenceInCalendarDays, format } from "date-fns";
-import moment from "moment";
 import Link from "next/link";
 import React, { useState } from "react";
 import {
-  FetchUserGoalsDataResponse,
-  FETCH_USER_GOALS,
   UserGoalsData,
 } from "../../graphql/fetchUserGoals";
-import { useAuth } from "../../lib/authContext";
-import { User } from "./profileV2/ProfileHeaderComponent";
-
 export type GoalsSectionProps = {
   sectionName?: string;
   userGoals?: UserGoalsData[];
--- a/components/coding/GoalsSectionComponent.tsx
+++ b/components/coding/GoalsSectionComponent.tsx
@@ -62,6 +62,9 @@ export default function GoalsSection({
             <p className="hidden md:block col-span-1 md:col-span-2">
               {format(new Date(goal.targetDate), "MM/dd/yyyy")}
             </p>
+            <p className="md:hidden col-span-1">
+              {format(new Date(goal.targetDate), "MM/dd")}
+            </p>
             <p className="hidden md:block md:col-span-2">
               {differenceInCalendarDays(new Date(goal.targetDate), new Date())}
             </p>
--- a/components/coding/GoalsSectionComponent.tsx
+++ b/components/coding/GoalsSectionComponent.tsx
@@ -59,7 +59,7 @@ export default function GoalsSection({
             <p className="hidden md:block md:col-span-2">
               {format(new Date(goal.createdAt), "MM/dd/yyyy")}
             </p>
-            <p className="md:col-span-2">
+            <p className="hidden md:block col-span-1 md:col-span-2">
               {format(new Date(goal.targetDate), "MM/dd/yyyy")}
             </p>
             <p className="hidden md:block md:col-span-2">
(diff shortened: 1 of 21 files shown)


iamnotreddy (changes requested):
- let's rewrite the returnGoalStyle function so that it takes in a goal of type UserGoalsData and move it outside the component definition... this way we can move it outside the map function (idk why I put it there in the first place) and we can use it across the repo if we so choose

- we can just call differenceInCalendarDays from date-fns directly, think it's overkill to define a separate function

- I don't think we actually need abridgedUserGoals. How about we load userGoals on the profile similar to [this](https://github.com/skillify-ca/skillify-web/blob/2e0042c298d7f711aee9d52a948699fbc1473227/pages/goals/index.tsx#L42-L65) but only load the "Current Section" and do the slicing here (so we only get the three most recent goals). This way our component is just rendering whatever data we supply it, not much logic going on inside

- let's keep using redux for this part since we already have it set up for goals

I don't think we break anything with this approach. Down the line, let's think about where / when we're making our Hasura calls because I'm seeing some inefficiency... currently if you go to the profile page, we query all your goals, then if you navigate to the goal page we query all your goals again. Don't need to solve for that in this PR but I think we should eventually

iamnotreddy:
One other note - I think we can handle showing fewer fields on the mobile view with tailwind [visibility classes](https://tailwindcss.com/docs/visibility) + [mobile breakpoints](https://tailwindcss.com/docs/screens)

kwiedenhaupt:
All changes requested were made/addressed. This is currently the mobile vs. PC view.
Screen Shot 2022-12-08 at 10 41 31 AM
Screen Shot 2022-12-08 at 10 41 44 AM

components/coding/GoalsSectionComponent.tsx
```diff
@@ -1,18 +1,32 @@
...
 import { UserGoalsData } from "../../graphql/fetchUserGoals";
-
 export type GoalsSectionProps = {
-  userGoals: UserGoalsData[];
-  sectionName: string;
+  sectionName?: string;
+  userGoals?: UserGoalsData[];
+};
+
+export const returnGoalStyle = (goal) => {
```
iamnotreddy, line 11: let's define the type for the function input so Typescript knows what's available on goal.. 

i.e const returnGoalStyle = (goal: UserGoalsData)...

pages/profile/v2/[slug].tsx
```diff
@@ -1,12 +1,15 @@
 import { ApolloClient, InMemoryCache } from "@apollo/client";
+import GoalsSectionComponent from "../../../components/coding/GoalsSectionComponent";
```
Comment (iamnotreddy) [inline pages/profile/v2/[slug].tsx:2]: not sure why your linter didn't pick up these blank spaces

components/coding/profileV2/BadgesDisplayedComponent.tsx
```diff
@@ -1,3 +1,4 @@
+
 import React from "react";
 
 import { User } from "./ProfileHeaderComponent";
```
iamnotreddy, line 4: You moved this to fetchUserProfile so need to update the import path

Is your linter broken? I feel like it should alert you to changed imports and remove the random blank rows across your code

components/coding/GoalsSectionComponent.tsx
```diff
@@ -27,44 +41,26 @@ export default function GoalsSection({
...
+              goal
             )}`}
           >
             <p>{index + 1}.</p>
             <p className="col-span-2 md:col-span-4">{goal.goalName}</p>
             <p className="hidden md:block md:col-span-2">
               {format(new Date(goal.createdAt), "MM/dd/yyyy")}
             </p>
-            <p className="md:col-span-2">
+            <p className="hidden md:block col-span-1 md:col-span-2">
```
iamnotreddy, line 55: nice!

components/coding/profileV2/ProfileHeaderComponent.tsx
```diff
@@ -16,6 +17,7 @@ import BadgesDisplayedComponent from "./BadgesDisplayedComponent";
 import JoinedDateComponent from "./JoinedDateComponent";
 
 export type UserProfileSectionProps = {
+
```
iamnotreddy, line 20: not sure why your linter didn't pick up these blank spaces

iamnotreddy (approved):
a few changes but otherwise looks good! approved for now but please fix these before merging

nice work

components/coding/GoalsSectionComponent.tsx
kwiedenhaupt: thank you! did that.

PR #826, opened 2022-12-12
Profile v2 assignment component beginning

kwiedenhaupt:
Pictures of open/closed states for the expandable container and then the rendering of the AssignmentSectionComponent on the profileV2 page.

Screen Shot 2022-12-11 at 6 53 07 PM
Screen Shot 2022-12-11 at 6 52 41 PM

1. Created a new AssignmentSectionComponent to work alongside the GoalsSectionComponent to be implemented on profileV2 page.
2. Created a new query to pull all assignments instead of just submissions for one particular assignment.
3. Implemented ExpandableContainer on both Goals and Assignments sections.


graphql/fetchAllUserAssignments.ts
```diff
@@ -0,0 +1,19 @@
+import { gql } from "@apollo/client";
+export const FETCH_ALL_USER_ASSIGNMENTS = gql`
```
iamnotreddy, line 2: let's rewrite this query so we get assignment name from the coding_assignments table - this should be possible since both tables have a relationship defined

components/coding/profileV2/AssignmentSectionComponent.tsx
```diff
@@ -0,0 +1,87 @@
...
+        },
+      }
+    );
+  return (
+    <ExpandableContainer open={true} title={""}>
+      <div>
+        {userAssignments.length > 0 && (
+          <div className="grid grid-cols-5 text-sm font-semibold text-center border-b-2 md:grid-cols-12 md:text-lg">
+            <p className="col-span-2 md:col-span-5">Assignment</p>
+            <p className="hidden md:block md:col-span-2">Completed</p>
```
iamnotreddy, line 44: how about we have one "status" column instead of two separate columns, then display the check mark if completed, and the x icon if waiting review... 

design nitpick, but wondering if there's a different icon we can use for whether an assignment is awaiting review

components/coding/profileV2/AssignmentSectionComponent.tsx
```diff
@@ -0,0 +1,87 @@
...
+                  )}
+                </p>
+                <p className="hidden md:block col-span-1 md:col-span-2">
+                  {assignment.review_link ? (
+                    <CheckCircleIcon className="w-5 h-5 cursor-pointer hover:text-yellow-600" />
+                  ) : (
+                    <XCircleIcon className="w-5 h-5 cursor-pointer hover:text-yellow-600" />
+                  )}
+                </p>
+                <Link href={"/web/React/assignments/template/"}>
```
iamnotreddy, line 77: i'd like to link to the assignment dynamically with a field from the coding_assignments table

iamnotreddy (changes requested):
I like what you did here - this looks clean! A few things to change before we merge this:
- let's have one "status" column (see below)
- let's modify the gql query so we can render the assignment name field instead of the id
- i'm thinking we add an assignment_link field to the coding_assignments table, so we have a path to reference dynamically on the edit button
- let's add an empty state if there are no assignment submissions - lmk if you need any thoughts on the design 

another thought - we could add another field to the table for the assignment submission (codesandbox) link... not sure what icon we'd use to stay consistent.

kwiedenhaupt:
Screen Shot 2022-12-12 at 12 37 26 PM
This is currently what the empty state looks like.

iamnotreddy (approved):
👍

PR #829, opened 2022-12-14
userGoals slice test

kwiedenhaupt:
1. created a unit test for the userGoalsSlice to test the setting of the state for goalsSections and userGoals.
2. tests pass when run.


iamnotreddy (approved):
everything looks good to me. nice work getting these tests to all pass!

PR #830, opened 2022-12-15
assignment component redux

kwiedenhaupt:
1. added a assignmentsSlice.ts
2. implemented the slice to AssignmentSectionComponent.tsx
seems to work as expected.


components/coding/profileV2/AssignmentSectionComponent.tsx
```diff
@@ -32,10 +37,34 @@ export default function AssignmentsSection({}: AssignmentSectionComponentProps)
         },
 
         onCompleted: (data: FetchUserAssignmentSubmissionsDataResponse) => {
-          setUserAssignments(data.user_assignment_submissions);
+          dispatch(setUserAssignments(data.user_assignment_submissions));
+          console.log("UA" + userAssignments);
         },
       }
     );
+
```
iamnotreddy, line 45: remove the commented code

iamnotreddy (approved):
Minus the cleanup, I think this is good and makes sense. Great work!

I spent some time thinking about the overall architecture of assignments. Some additional thoughts:
- we will need to replace local state in the assignment template - can do this in a separate PR
- I think having all assignments for a given user stored in this redux state like we have can work across the app
- If we just fetch all in-flight assignments for a user and store that in global state, that would eliminate the need for querying status for a specific assignment (which we're currently doing inside the assignment template)
- I think we will need to create another reducer to update assignment status i.e when a user submit's their assignment

PR #834, opened 2022-12-16
created a fix for if there is a long title

kwiedenhaupt:
1. created a parent div that sets the width and then on the child <p> the text truncates if it's longer than the set width
2. hover is set to show the entire title
3. this does not apply if over 8 characters

also readjusted columns so that they are better aligned
Screen Shot 2022-12-14 at 3 57 09 PM
(1. new styling for mobile view)
Screen Shot 2022-12-14 at 3 55 05 PM
(2. this is the view if the assignment title is very long -- truncated)
Screen Shot 2022-12-14 at 3 54 54 PM
(3. this is the view when you hover over the title -- more is shown but it still cuts off when it goes beyond a certain point)
Screen Shot 2022-12-14 at 3 56 50 PM
(4. this is the current view with the "template" shorter assignment name in full screen view)

### DIFF
--- a/components/coding/profileV2/AssignmentSectionComponent.tsx
+++ b/components/coding/profileV2/AssignmentSectionComponent.tsx
@@ -16,6 +16,30 @@ import {
 import { useAuth } from "../../../lib/authContext";
 import ExpandableContainer from "../ExpandableContainer";
 
+export const returnWrapStyling = (
+  assignment: UserAssignmentSubmissionsData
+) => {
+  let wrapStyle = "";
+  if (assignment.coding_assignment.assignment_name.length >= 3) {
+    wrapStyle = "truncate";
+  } else {
+    null;
+  }
+  return wrapStyle;
+};
+
+export const returnParentStyling = (
+  assignment: UserAssignmentSubmissionsData
+) => {
+  let parentStyle = "";
+  if (assignment.coding_assignment.assignment_name.length >= 8) {
+    parentStyle = "w-[100px] hover:w-full";
+  } else {
+    null;
+  }
+  return parentStyle;
+};
+
 export type AssignmentSectionComponentProps = {};
 
 export default function AssignmentsSection({}: AssignmentSectionComponentProps) {
@@ -36,13 +60,14 @@ export default function AssignmentsSection({}: AssignmentSectionComponentProps) 
         },
       }
     );
+
   return (
     <ExpandableContainer open={true} title={""}>
       <div>
         {userAssignments.length > 0 && (
-          <div className="grid grid-cols-5 text-sm font-semibold text-center border-b-2 md:grid-cols-12 md:text-lg">
-            <p className="col-span-2 md:col-span-6">Assignment</p>
-            <p className="col-span-2 font-semibold md:col-span-2">Status</p>
+          <div className="grid grid-cols-9 text-sm font-semibold text-center border-b-2 md:grid-cols-12 md:text-lg">
+            <p className="col-span-4 md:col-span-6">Assignment</p>
+            <p className="col-span-3 font-semibold md:col-span-2">Status</p>
           </div>
         )}
 
@@ -56,12 +81,18 @@ export default function AssignmentsSection({}: AssignmentSectionComponentProps) 
           userAssignments.map((assignment, index) => {
             return (
               <div
-                className={`grid grid-cols-5 my-2 text-sm text-center md:grid-cols-12 md:text-lg place-items-center ${assignment}`}
+                className={`grid grid-cols-7 my-2 text-sm text-left md:grid-cols-12 md:text-lg md:place-items-center  ${assignment}`}
               >
-                <p className="col-span-1">{index + 1}.</p>
-                <p className="col-span-1 md:col-span-4">
-                  {assignment.coding_assignment.assignment_name}
-                </p>
+                <p className="place-item-left col-span-1">{index + 1}.</p>
+                <div
+                  className={`col-span-3 md:col-span-4 ${returnParentStyling(
+                    assignment
+                  )}`}
+                >
+                  <p className={` ${returnWrapStyling(assignment)}`}>
+                    {assignment.coding_assignment.assignment_name}
+                  </p>
+                </div>
                 <p className="col-span-2 md:block md:col-span-4">
                   {assignment.submission_link ? (
                     assignment.review_link === null ? (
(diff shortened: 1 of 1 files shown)


components/coding/profileV2/AssignmentSectionComponent.tsx
```diff
@@ -16,6 +16,30 @@ import {
 import { useAuth } from "../../../lib/authContext";
 import ExpandableContainer from "../ExpandableContainer";
 
+export const returnWrapStyling = (
+  assignment: UserAssignmentSubmissionsData
+) => {
+  let wrapStyle = "";
+  if (assignment.coding_assignment.assignment_name.length >= 3) {
+    wrapStyle = "truncate";
+  } else {
```
iamnotreddy, line 25: no need for this else clause; if the if statement returns false, it will return an empty string wrapStyle

components/coding/profileV2/AssignmentSectionComponent.tsx
```diff
@@ -16,6 +16,30 @@ import {
...
+  return wrapStyle;
+};
+
+export const returnParentStyling = (
+  assignment: UserAssignmentSubmissionsData
+) => {
+  let parentStyle = "";
+  if (assignment.coding_assignment.assignment_name.length >= 8) {
+    parentStyle = "w-[100px] hover:w-full";
+  } else {
```
iamnotreddy, line 37: same thing here

components/coding/profileV2/AssignmentSectionComponent.tsx
```diff
@@ -16,6 +16,30 @@ import {
...
+  }
+  return wrapStyle;
+};
+
+export const returnParentStyling = (
+  assignment: UserAssignmentSubmissionsData
+) => {
+  let parentStyle = "";
+  if (assignment.coding_assignment.assignment_name.length >= 8) {
+    parentStyle = "w-[100px] hover:w-full";
```
iamnotreddy, line 36: haven't seen the syntax of 100px wrapped in brackets - what does that do? i try to avoid using absolute sizes (100) as opposed to relative (1/2, 1/3, full etc.) as it can cause unintended issues with mobile responsiveness

iamnotreddy: shouldn't have "hover:w-full" over here... when I hover over the assignment name, the name jumps to the left side of the table.

were you trying to do something different here?

components/coding/profileV2/AssignmentSectionComponent.tsx
```diff
@@ -16,6 +16,30 @@ import {
 import { useAuth } from "../../../lib/authContext";
 import ExpandableContainer from "../ExpandableContainer";
 
+export const returnWrapStyling = (
```
iamnotreddy, line 19: something to think about here - do other files in the app need to use these functions? if so we should "export" them so they are available all over our app.. if not we can just keep them local (i.e remove the export)

in this case, the function takes in an assignment-specific data type and returns styling specific to this component, so I think it makes sense to just contain it to this file

if we did plan on making this function more generic, i.e use it to style any div, and if we have several other functions like this, I'd say we should create a separate helper functions file and export it from that file

components/coding/profileV2/AssignmentSectionComponent.tsx
iamnotreddy: this is what it does when i hover over: 

Screen Shot 2022-12-19 at 2 48 18 PM

components/coding/profileV2/AssignmentSectionComponent.tsx
```diff
@@ -56,12 +81,18 @@ export default function AssignmentsSection({}: AssignmentSectionComponentProps)
...
             return (
               <div
-                className={`grid grid-cols-5 my-2 text-sm text-center md:grid-cols-12 md:text-lg place-items-center ${assignment}`}
+                className={`grid grid-cols-7 my-2 text-sm text-left md:grid-cols-12 md:text-lg md:place-items-center  ${assignment}`}
               >
-                <p className="col-span-1">{index + 1}.</p>
-                <p className="col-span-1 md:col-span-4">
-                  {assignment.coding_assignment.assignment_name}
-                </p>
+                <p className="place-item-left col-span-1">{index + 1}.</p>
```
iamnotreddy, line 86: what does place-item-left do? i'm not 100% sure it exists in tailwind, tried looking for it in the documentation.. lmk if you disagree

https://tailwindcss.com/docs/place-items#stretch

iamnotreddy (changes requested):
left a few comments here on things we need to rework, lmk if you want to discuss when we chat tomorrow

i think in general with tailwind, it's important to know the purpose of each piece of code, because it gives us no errors if something doesn't work 

feel like this is a big drawback to tailwind - our code can become messy and unreadable if we aren't careful. this is something i'm working on because my early tailwind code was super unreadable and i want to avoid that!

components/coding/profileV2/AssignmentSectionComponent.tsx
kwiedenhaupt: great catch.  removed!

components/coding/profileV2/AssignmentSectionComponent.tsx
kwiedenhaupt: removed!

components/coding/profileV2/AssignmentSectionComponent.tsx
kwiedenhaupt: it was a mistake -- was supposed to be place-items-start -- but now that I removed it, the change is so negligible that I don't think it's worth it and took it out!

components/coding/profileV2/AssignmentSectionComponent.tsx
kwiedenhaupt: I was attempting to show the entire title if it's long on hover -- but you're right that it doesn't make sense for it to jump left if it's short.  I think just the truncate on mobile will be sufficient.  I think it's best to remove the hover also.

components/coding/profileV2/AssignmentSectionComponent.tsx
kwiedenhaupt: you're right! removed the export because the files need only to be used locally.

components/coding/profileV2/AssignmentSectionComponent.tsx
```diff
@@ -21,6 +21,21 @@ import {
...
+  }
+  return wrapStyle;
+};
+
+const returnParentStyling = (assignment: UserAssignmentSubmissionsData) => {
+  let parentStyle = "";
+  if (assignment.coding_assignment.assignment_name.length >= 8) {
+    parentStyle = "w-1/8";
+  } else return parentStyle;
+};
```
iamnotreddy, line 37: if the condition is true this sets the parentStyle to 'w-1/8' but doesn't return anything; if it's false it will return a blank string

you should set this function to always return parentStyle - in the case the condition is true, it will return a parentStyle of "w-1/8"

you're doing this correctly with the returnWrapStyling function

iamnotreddy (approved):
thanks for making the changes! 

approved but please fix your returnParentStyling function

PR #835, opened 2022-12-18
fix four small bugs on goals

kwiedenhaupt:
1. attempted fix to bug where choosing the date was sometimes one day off.  The fix is implemented on this page.  You can seemingly now choose the date you want and not the day after.
Screen Shot 2022-12-18 at 8 47 20 AM
2. put in a check that does not allow you to edit a goal with a target date that is already past. (alert message appears as below)
Screen Shot 2022-12-17 at 5 11 25 PM
3. Fixed the bug where there was an empty ExpandableContainer if any of the sectionName such as "archived" or "completed" or "current" were empty.  
SCREENSHOT of BUG
Screen Shot 2022-12-17 at 5 39 41 PM
SCREENSHOT of FIX
Screen Shot 2022-12-17 at 5 12 49 PM
4. Implemented a "No Goals" section on profileV2 page in case the user has no goals in Hasura.
Screen Shot 2022-12-17 at 5 57 47 PM


pages/goals/[slug].tsx
```diff
@@ -129,15 +129,16 @@ const EditGoalsPage = () => {
             <input
               type="date"
               className="w-1/2 p-2 text-left border rounded-md shadow-md md:w-1/4 text-murkrow"
-              value={format(
-                new Date(editedGoalValues.targetDate),
-                "yyyy-MM-dd"
-              )}
+              value=""
```
Comment (iamnotreddy) [inline pages/goals/[slug].tsx:132]: having an empty string here means the date will never actually update: 

Screen Shot 2022-12-19 at 3 08 00 PM

pages/goals/[slug].tsx
```diff
@@ -129,15 +129,16 @@ const EditGoalsPage = () => {
...
               onChange={(e) => {
-                setEditedGoalValues((prevState) => ({
-                  ...prevState,
-                  targetDate: new Date(e.target.value + "T00:00:00"),
-                }));
+                if (new Date(e.target.value + "T00:00:00") > new Date()) {
```
Comment (iamnotreddy) [inline pages/goals/[slug].tsx:134]: nice add!

pages/goals/addGoal.tsx
```diff
@@ -59,7 +59,7 @@ const EditGoalsPage = () => {
           onChange={(e) => {
             setNewGoalValues((prevState) => ({
               ...prevState,
-              targetDate: new Date(e.target.value),
+              targetDate: new Date(e.target.value + "T00:00:00"),
```
iamnotreddy, line 62: 🎉

components/coding/GoalsSectionComponent.tsx
```diff
@@ -74,5 +76,12 @@ export default function GoalsSection({
         })}
       </div>
     </ExpandableContainer>
-  );
+  ) : userGoals.length <= 0 && profile === true ? (
+    <ExpandableContainer open={true} title={""}>
+      <div className="col-span-3 p-8 mb-8 text-center shadow-md bg-slate-300 dark:bg-slate-900">
+        Click on the "Goals" tab on the sidebar to get ahead by creating and
```
iamnotreddy, line 82: love accounting for the empty state!

components/coding/GoalsSectionComponent.tsx
```diff
@@ -7,6 +7,7 @@ import ExpandableContainer from "./ExpandableContainer";
 export type GoalsSectionProps = {
   sectionName?: string;
   userGoals?: UserGoalsData[];
+  profile?: boolean;
```
iamnotreddy, line 10: Can you explain how we're using this new type? I think I get what you're doing here - if we're calling this component on the profile, set to true, set to false if not. But I'm not sure I see where it's actually being used?

nitpick on naming - I'd probably change the name so it's more apparent it's a boolean. Maybe like onProfile or inProfile

iamnotreddy (changes requested):
question re: the new type and need to fix the the value field inside the date input box

thanks for taking initiative to fix all these bugs, very thoughtful and will make the user experience much smoother :)

pages/goals/[slug].tsx
Comment (kwiedenhaupt) [reply pages/goals/[slug].tsx]: great catch! that was done by accident.. put the following code back in :

`value={format(
                new Date(editedGoalValues.targetDate),
                "yyyy-MM-dd"
              )}`

components/coding/GoalsSectionComponent.tsx
kwiedenhaupt: I'm only using the new type here because we are using the GoalsSectionComponent right now for the goals index page and on the profile index page v2.  This is a problem because if there are no userGoals for the user on the profile page, I want to display the box saying there are no user goals.  If there are no userGoals in a section on the goals index page, like "archived" or "completed" or "current" then the same box will display and that looks bad I think. If the user has no goals it's better to see this I think: 
Screen Shot 2022-12-19 at 5 46 06 PM

Maybe this would be better?

iamnotreddy: I think either the inProfile or enum approach both work - I'd probably lean towards the inProfile boolean because it's simpler and I don't anticipate us placing this in too many other places? If we needed to place it on more than two different pages and specify different behavior for each page, might make sense to use an enum

iamnotreddy (approved):

components/coding/GoalsSectionComponent.tsx
kwiedenhaupt: changed to inProfile.  What you're saying makes sense.

PR #838, opened 2022-12-20
Setup Achievement Component 

LuckyHariharan:
In this PR I setup the Achievement component under profile/v2 in an expandable container component with the following: 
- Passing the user badge data as a prop to the achievement component
- Adding an edit Icon at each badge section that currently links to the goals page 
image

@raveenreddy0 please feel free to tear this PR up! Let me know what changes you'd like to see made.


iamnotreddy (approved):
just fix the edit button so it only appears once and this is good

thinking edit button in the top right corner

PR #840, opened 2022-12-27
Changes that needed to be committed from PR 835

kwiedenhaupt:
1. Changing the enum to inProfile and then changing associated index files.


components/coding/GoalsSectionComponent.tsx
```diff
@@ -80,14 +80,12 @@ export default function GoalsSection({
         })}
       </div>
     </ExpandableContainer>
-  ) : userGoals.length <= 0 &&
-    componentUsageType === GoalsComponentSection.PROFILE ? (
+  ) : userGoals.length <= 0 && inProfile === true ? (
```
iamnotreddy, line 83: tip: when referencing a boolean, the default is always true so you can just write "inProfile" - does the same thing as your code above

pages/profile/v2/index.tsx
```diff
@@ -54,7 +52,7 @@ export default function Profile(props) {
 
       <div className="grid grid-cols-1">
         <GoalsSectionComponent
-          componentUsageType={GoalsComponentSection.PROFILE}
+          inProfile={true}
```
iamnotreddy, line 55: same thing here, you could just write inProfile

pages/goals/index.tsx
```diff
@@ -83,7 +81,7 @@ export default function Goals(props) {
               <GoalsSectionComponent
                 userGoals={section.userGoals}
                 sectionName={section.sectionName}
-                componentUsageType={GoalsComponentSection.GOALS}
+                inProfile={false}
```
iamnotreddy, line 84: over here, you could write !inProfile for shorthand to pass the value as false into the component

iamnotreddy (approved):
approved!

you don't have to change this but one piece of feedback - I think the use of ternary operators in your rendering logic is a bit confusing, especially since you're evaluating more than two conditions... in these kinds of cases, i think it's better for readability to use an if statement

kwiedenhaupt:
Screen Shot 2022-12-29 at 4 37 34 PM
Brand new approach focused on perfecting the query and then dispatching data, and it seems to be working!

PR #843, opened 2022-12-31
skill ratings functions unit tests

kwiedenhaupt:
Screen Shot 2022-12-30 at 7 36 04 PM
Created three tests for transformSkillsAndRatings function to ensure inputs into the function are as expected and are transformed correctly so they can be saved to redux and deployed in the feature.  This will ensure proper functionality of the component.
Case 1:  All studentRatings are in DB
Case 2: Some studentRatings are in the DB
Case 3: No studentRatings are in the DB (randomly generated numbers in place of IDs)

Created 1 test covering two scenarios for transformSkillsForDB function to ensure skills can be saved to the DB and will not crash the feature.
Case 1: A userSkillId is generated by Math.random() and cannot be saved to the DB and should not be returned
Case 2: A userSkillId was generated by the DB and should be used to write back to the DB


redux/skillRatingsSlice.ts
```diff
@@ -6,7 +6,7 @@ export type SkillRatingsState = {
 };
 
 export type SkillRatingsRow = {
-  userSkillId: String;
+  userSkillId: String|number;
```
vithushan19, line 9: Why is this needed?

pages/api/skillRatingsFunctions.ts
```diff
@@ -0,0 +1,56 @@
...
+        : 0,
+        userSkillId: row.intro_course_skills_users[0]?.id ? row.intro_course_skills_users[0].id : Math.random(),
+      };
+    });
+  };
+
+  export const transformSkillRatingForDB = (skillRatings: SkillRatingsRow[], user:User) => {
+    // map from redux type to write back to DB
+    const transformedOutput = skillRatings.map((row) => {
+      if (typeof row.userSkillId == "number") {
```
vithushan19, line 39: I think we can ignore this case. User IDs should probably always be strings.

If they are numbers in the database that is because they created their user account before I migrated our auth system. All new users will have strings

pages/api/skillRatingsFunctions.ts
```diff
@@ -0,0 +1,56 @@
...
+        skillName: row.intro_course_skill["name"],
+        unitName: row.intro_course_skill["intro_course_unit"]["title"],
+        studentRating: parseInt(row.studentRating),
+      };
+    });
+
+    return mappedSkillRating;
+  };
+
+  export const transformSkillsAndRatings = (skillRatings: SkillsAndRatings[]) => {
```
vithushan19, line 21: Nice!

vithushan19 (commented):
Looks good to me

vithushan19:
Nice work adding tests

redux/skillRatingsSlice.ts
kwiedenhaupt: I generated a random number between 0 and 1 and assigned it as a "placeholder" userSkillId in the transformSkillsAndRatings function.  This is to account for the number type of that function. The userSkillId is the id for the user's particular assignment rating.  When I am displaying the ratings initially in the transformSkillsAndRatings function, if students don't have a rating for a particular skill I need to set that rating to 0.  But in this case those id's have not yet been generated by hasura and need to be created in this function through that "placeholder" userSkillId so that the slider bars work (rating is attached to a particular skill) on the display.  If a user changes these ratings on the page through using the slider bar and I want to save them back to hasura, I cannot save the "placeholder" id.  I have to skip the userSkillId (randomly generated number placeholder) in the transformSkillRatingsForDB function.  I identify whether the userSkillId has been created by hasura or needs to be skipped through checking if it's a number or not in the transformSkillRatingsForDB function.  

This number type accounts for the randomly generated userSkillIds.

Does this make sense?

pages/api/skillRatingsFunctions.ts
kwiedenhaupt: see comment above.

iamnotreddy (approved):
these tests look good! feel like you're thoroughly testing this function with the three use cases we've outlined

make sure you remove any changes that aren't from the test file - you should be able to drop those other files

pages/api/skillRatingsFunctions.test.ts
```diff
@@ -0,0 +1,255 @@
...
+            },
+            "intro_course_skills_users_aggregate": {
+              "nodes": []
+            }
+          },
+    ];
+
+
+   // Mocking Math.random to always return 0.5
+const mockMath = Object.create(global.Math);
```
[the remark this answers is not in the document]
kwiedenhaupt: I believe it's run in the test and it automatically overwrites the Math.random number generated by the function to be 0.5 globally in this file only until it is turned off

PR #844, opened 2023-01-03
Achievement Component Transform Data Type and Refactor

LuckyHariharan:
In this PR the data has three levels of nested types that has been transformed into one level of nested type. The achievement component was refactored to display using this new transformed data. 

Next step is to create a save button which uses the transformed data to insert or delete a row in the user coding badges table in Hasura.


components/coding/profileV2/AcheivementComponent.tsx
```diff
@@ -1,62 +1,132 @@
...
+        };
+      });
 
-const AcheivementComponent = ({ user, data }) => {
-  const [units, setUnits] = useState<IntroCourseUnit[]>([]);
+      return updatedData;
+    });
+  };
+
+  const transformData = (data: Data) => {
```
iamnotreddy, line 50: two things: 
- naming here is pretty generic, could end up being confusing down the line
- might be nice to move this to a helpers file since it doesn't have to be in here, in which case i'd definitely make the name more descriptive

components/coding/profileV2/AcheivementComponent.tsx
```diff
@@ -1,62 +1,132 @@
...
+          };
+        }),
+      };
+    });
+    return transformedOutput;
+  };
   useEffect(() => {
-    if (typeof data !== "undefined") {
-      setUnits(data.intro_course_unit);
+    if (data != undefined) {
```
iamnotreddy, line 67: `if (data)` is shorthand for checking whether the object is defined

components/coding/profileV2/AcheivementComponent.tsx
```diff
@@ -1,62 +1,132 @@
 import React, { useEffect, useState } from "react";
 import ExpandableContainer from "../ExpandableContainer";
 import { PencilAltIcon } from "@heroicons/react/outline";
-import { IntroCourseUnit } from "../../../graphql/coding/userBadges/fetchUserBadges";
-import Link from "next/link";
+import { Data } from "../../../graphql/coding/userBadges/fetchUserBadges";
 import { User } from "../../../graphql/fetchUserProfile";
+import _ from "lodash";
```
iamnotreddy, line 6: remove this import since we're not using it (and generally try to remove all other unused imports)

iamnotreddy (approved):
great work putting this together! left a few suggestions

if you're set on using the original table structure of user_coding_badges like we discussed yesterday, might be a good idea to close this PR out and start the gql work in another PR

LuckyHariharan:
> great work putting this together! left a few suggestions
> 
> if you're set on using the original table structure of user_coding_badges like we discussed yesterday, might be a good idea to close this PR out and start the gql work in another PR

Sounds good! Thanks again for your feedback Raveen!

PR #845, opened 2023-01-04
returnGoalStyle mock date modifications

kwiedenhaupt:
Added 2 tests:
1.  yellow highlight where the days remaining is 3 or less.
2. no highlight where the days remaining is greater than 3

Added Mock Date package and modified package.JSON
In order to do this, the only way I could get it to work was to add a mock date library and then add a line in the script file in jest regarding setting the time zone to UTC in the package.JSON file.
Screen Shot 2023-01-03 at 8 05 15 PM


package.json
```diff
@@ -101,6 +101,7 @@
     "eslint-config-prettier": "^8.3.0",
     "eslint-plugin-react": "^7.26.1",
     "jest": "^26.6.3",
+    "mockdate": "^3.0.5",
```
vithushan19, line 104: I'm ok with adding mockdate. Adding a dev dependency isn't as bad as adding a regular dependency.

I believe that dev dependencies won't be included in production builds only local debug builds.

components/coding/GoalsSectionComponent.test.ts
```diff
@@ -1,38 +1,102 @@
...
+
+  // Act
+  const result = returnGoalStyle(goal);
+
+  // Assert
+  expect(result).toBe(yellowHighlightStyle);
+
+});
+
+test("Test returnGoalStyle function - no highlight for a further future target date", async () => {
```
vithushan19, line 58: Awesome!

vithushan19 (approved):
This is great! You're really thinking about code in a different way.

PR #846, opened 2023-01-06
Tabbed UI feature

kwiedenhaupt:
1. removed SkillSections component usage and filtered the data to obtain a list of unique section names that I used to map the sections out to render in the component.
2. created buttons labeled appropriately for each section
3. updated the UI to look like buttons in asana
4. moved setting the sections and the active tab into the useEffect so that I could have a default dynamic value in the state hook for the activeTab.  The useEffect was needed to in order for the sections variable to populate before the component renders so that skills show up upon loading.

### DIFF
--- a/components/coding/SkillRatingsComponent.tsx
+++ b/components/coding/SkillRatingsComponent.tsx
@@ -60,14 +60,29 @@ export default function SkillRatingsComponent(props) {
     return <div>Loading...</div>;
   }
 
+  const setActiveTabStyling = (tab: string) => {
+    let styling = "";
+    if (tab === activeTab) {
+      styling =
+        "ml-8 justify-content-center text-2xl text-gray-500 w-36 py-2 h-12 text-black-500 underline decoration-[0.18rem] underline-offset-[16px]";
+    } else {
+      styling =
+        "ml-8 justify-content-center text-2xl text-gray-500 w-36 py-2 h-12 cursor-pointer hover:text-black-500 hover:underline hover:hover:decoration-[0.18rem] hover:underline-offset-[16px]";
+    }
+    return styling;
+  };
+
   return (
     <div className="flex flex-row overflow-auto-bg-scroll">
       <ExpandableContainer open={true} title={""}>
         <div className="space-x-10">
           {sections.map((it) => (
             <button
-              className="ml-8 justify-content-center text-2xl text-gray-500 w-36 py-2 h-12 cursor-pointer hover:text-black-500 hover:underline hover:hover:decoration-[0.18rem]"
-              onClick={() => setActiveTab(it)}
+              className={setActiveTabStyling(it)}
+              onClick={() => {
+                setActiveTab(it);
+                setActiveTabStyling(it);
+              }}
             >
               {it}
             </button>
@@ -105,3 +120,5 @@ export default function SkillRatingsComponent(props) {
     </div>
   );
 }
+
+//"ml-8 justify-content-center text-2xl text-gray-500 w-36 py-2 h-12 cursor-pointer hover:text-black-500 hover:underline hover:hover:decoration-[0.18rem] hover:underline-offset-[16px]"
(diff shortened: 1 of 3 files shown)


kwiedenhaupt:
I wasn't sure about text sizing -- it didn't quite look right for the section tabs to be this size, larger or smaller, open to suggestions about what looks the best.

kwiedenhaupt:
Screen Shot 2023-01-06 at 3 13 46 PM
adjusted underline offset and made it so that the active tab is always highlighted.

components/coding/SkillRatingsComponent.tsx
```diff
@@ -42,11 +45,64 @@ export default function SkillRatingsComponent(props) {
...
   });
 
+  useEffect(() => {
+    if (skillRatings.length > 0) {
+      const unitNames = skillRatings.map((skill) => skill.unitName);
+      const sections = unitNames.filter(
+        (unitName, index, array) => array.indexOf(unitName) === index
+      );
+      setSections(sections);
+      setActiveTab(sections[0]);
```
iamnotreddy, line 55: if you switch to the second tab and start changing the sliders, it automatically jumps back to the first tab because of this line... skillRatings is in the dependency array of this useEffect, so the hook runs any time its value changes

i think you could actually move this logic into the onCompleted of your query that fetches the skill ratings

iamnotreddy (changes requested):
the UI looks great - nice work! you'll need to fix the one comment so skill ratings can be set correctly on other tabs

components/coding/SkillRatingsComponent.tsx
```diff
@@ -42,11 +45,64 @@ export default function SkillRatingsComponent(props) {
...
+    }
+  }, [skillRatings]);
+
+  if (activeTab === "") {
+    return <div>Loading...</div>;
+  }
+
+  const setActiveTabStyling = (tab: string) => {
```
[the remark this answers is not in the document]
kwiedenhaupt: good idea. changed to activeTabStyling

components/coding/SkillRatingsComponent.tsx
```diff
@@ -42,11 +45,64 @@ export default function SkillRatingsComponent(props) {
...
+    let styling = "";
+    if (tab === activeTab) {
+      styling =
+        "ml-8 justify-content-center text-2xl text-gray-500 w-36 py-2 h-12 text-black-500 underline decoration-[0.18rem] underline-offset-[18px]";
```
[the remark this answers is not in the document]
kwiedenhaupt: great idea! updated.

components/coding/SkillRatingsComponent.tsx
kwiedenhaupt: That doesn't seem to work, because the useEffect is needed in order to make sure skillRatings is populated before the sections are set.  I was able to get it to work by creating a hasRun state hook.  This will be set to false initially, and then the useEffect will check !hasRun and if it's false will run the useEffect hook logic which at the end will set the hasRun to true -- which will cause the useEffect to stop running when skillRatings changes.  

The skillRatings dependency is needed to retrigger a run, because it needs to run multiple times at the beginning to keep checking for a value in order to initially set the skillRatings.  Otherwise, I'm stuck with a "loading..." screen.

The useEffect is needed in the first place because if the logic is put in the onComplete, it will try to set the sections always before a skillRatings value is set, and I will get nothing -- only a save button. 

There def could be a better solution over adding the hasRun -- so I'm open to suggestions!

iamnotreddy: I think this solution is pretty good actually! Nice thought process through this - I think you're getting a better grasp on the order of operations. 

There might be a way to incorporate a Promise in the onCompleted to work around the issues you mentioned but this definitely works and is likely less complicated.

I'd maybe call `hasRun` something more descriptive like `haveTabsLoaded` so it's more clear what it does.

iamnotreddy (approved):
🥳

components/coding/SkillRatingsComponent.tsx
kwiedenhaupt: perfect!

PR #847, opened 2023-01-06
fix bug for skillRatings when user has no skills in DB

kwiedenhaupt:
this should fix the bug. it was pulling the userSkillIds already in the database and therefore the upsert query didn't work and would just modify the records that already existed.

Now when the upsert query is called all records appear in the database for that user.
Screen Shot 2023-01-06 at 2 13 38 PM
partial screenshot of database populated with new userId records as desired:
Screen Shot 2023-01-06 at 2 17 22 PM


iamnotreddy (approved):
nice fix!

PR #848, opened 2023-01-08
Redesign expandable container closer to figma sketches

kwiedenhaupt:
1. added "title" in the expandable container and removed the section header from the individual components.  this allowed me to use a flex flex-row to display the plus/minus icon next to the title.
2. added border and shadowing around the container.
3. adjusted the shadow depending on if it was open or closed.
4. cleaned up some styling and unused tailwind code.
Screen Shot 2023-01-08 at 8 36 16 AM

### DIFF
--- a/components/coding/ExpandableContainer.tsx
+++ b/components/coding/ExpandableContainer.tsx
@@ -39,7 +39,7 @@ const ExpandableContainer: React.FC<ExpandableContainerProps> = ({
                 <MinusCircleIcon className="w-6 h-6 cursor-pointer hover:text-yellow-600" />
               )}
             </button>
-            <h6 className="text-3xl font-bold pb-2 md:text-4xl mt-3">
+            <h6 className="text-2xl font-bold pb-2 md:text-3xl mt-3">
               {title}
             </h6>
           </div>
--- a/components/coding/ExpandableContainer.tsx
+++ b/components/coding/ExpandableContainer.tsx
@@ -21,9 +21,13 @@ const ExpandableContainer: React.FC<ExpandableContainerProps> = ({
 
   return (
     <>
-      <div>
+      <div
+        className={`border-4 border-gray-900 rounded-lg ${
+          isOpen ? "shadow-lg shadow-gray-900" : ""
+        } `}
+      >
         <div>
-          <div className="p-2 border-bottom flex float left">
+          <div className="flex flex-row items-center p-4 space-x-2">
             <button
               className="mr-2"
               type="button"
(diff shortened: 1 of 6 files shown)


components/coding/GoalsSectionComponent.tsx
```diff
@@ -77,7 +77,7 @@ export default function GoalsSection({
       </div>
     </ExpandableContainer>
   ) : userGoals.length <= 0 && inProfile ? (
-    <ExpandableContainer open={true} title={""}>
+    <ExpandableContainer open={true} title={"Goals"}>
```
iamnotreddy, line 80: we should call the ExpandableContainer on the index page instead of within the component for all components - this way, each component is reusable across the app with or without the ExpandableContainer

this implementation is affecting the layout of the goal page:

Screen Shot 2023-01-09 at 2 19 23 PM

components/coding/ExpandableContainer.tsx
```diff
@@ -35,7 +39,9 @@ const ExpandableContainer: React.FC<ExpandableContainerProps> = ({
                 <MinusCircleIcon className="w-6 h-6 cursor-pointer hover:text-yellow-600" />
               )}
             </button>
-            <h6 className="">{title}</h6>
+            <h6 className="text-3xl font-bold pb-2 md:text-4xl mt-3">
```
iamnotreddy, line 42: design nit: text-3xl looks a little awkward with the profile header font, maybe something less than this?

Screen Shot 2023-01-09 at 2 22 24 PM

iamnotreddy (changes requested):
UI changes look good! Let's clean up the ExpandableContainer implementation so the goals page is unaffected

components/coding/GoalsSectionComponent.tsx
kwiedenhaupt: removed the container from all the individual components and added them all to the profile page.  Great suggestion!

components/coding/ExpandableContainer.tsx
kwiedenhaupt: reduced the title size by one...looks a lot better.

pages/profile/v2/index.tsx
```diff
@@ -32,30 +33,31 @@ export default function Profile(props) {
   return (
     <div className="flex flex-col p-4 m-4 overflow-auto bg-scroll">
       <ProfileHeaderComponent userId={user.uid} />
-
-      <h2 className="text-lg font-bold mt-14 mb-9">Assignments</h2>
-
-      <div className="grid grid-cols-1 mb-4">
-        <AssignmentSectionComponent />
+      <div className="grid grid-cols-1 my-9">
```
iamnotreddy, line 36: you can remove this 1 column grid from all these lines, i'd also recommend removing my-9 and using space-y-[] in the parent div

pages/profile/v2/index.tsx
```diff
@@ -32,30 +33,31 @@ export default function Profile(props) {
   return (
     <div className="flex flex-col p-4 m-4 overflow-auto bg-scroll">
```
iamnotreddy, line 34: space-y-[] here to space the expandable rows

iamnotreddy (approved):
looks much better! i'd clean up the HTML before merging

pages/profile/v2/index.tsx
kwiedenhaupt: perfect

PR #849, opened 2023-01-10
bypassing additional query if userAssignments global state exists

kwiedenhaupt:
1.  Added logic that checks to see if the userAssignments exists and if it does, it sets stages from the redux state.  If not, then the useLazyQuery fetchUserAssignmentSubmissions is run and the stages are set from there.
2. the link field to the incompleteStage set to "" was also added to satisfy the type

### DIFF
--- a/components/coding/studentPortal/AssignmentComponent.tsx
+++ b/components/coding/studentPortal/AssignmentComponent.tsx
@@ -114,11 +114,14 @@ export default function AssignmentComponent({
     return (
       <div>
         <p className="text-lg">{data.title}</p>
-        {data.screenshot ? (
-          <img src={data.screenshot} className="object-cover w-64 h-32 mb-4" />
+        {data.screenshotOrVideoId.kind === "screenshot" ? (
+          <img
+            src={data.screenshotOrVideoId.value}
+            className="object-cover w-64 h-32 mb-4"
+          />
         ) : (
           <iframe
-            src={`https://www.loom.com/embed/${data.videoId}`}
+            src={`https://www.loom.com/embed/${data.screenshotOrVideoId.value}`}
             frameBorder="0"
             webkit-allowfullscreen
             moz-allowfullscreen
--- a/components/coding/studentPortal/AssignmentComponent.tsx
+++ b/components/coding/studentPortal/AssignmentComponent.tsx
@@ -71,27 +71,32 @@ export default function AssignmentComponent({
 }: AssignmentComponentProps) {
   const [isOpen, setIsOpen] = useState(false);
 
+  const activeSectionStyling = () => {
+    let styling = "font-thin text-2xl";
+    if (!isOpen) {
+      styling = styling + " text-black-500 hover:text-slate-500";
+    } else {
+      styling = styling + " text-slate-500 hover:text-black-500";
+    }
+    return styling;
+  };
+
   if (data.component === "title") {
-    return <h1 className="text-5xl font-bold">{data.text}</h1>;
+    return <h1 className="text-5xl font-thin">{data.text}</h1>;
   } else if (data.component === "prompt") {
     return (
       <>
-        <div
-          className="flex p-4 cursor-pointer  w-1/6"
-          onClick={() => setIsOpen(!isOpen)}
-        >
-          <h1 className="font-bold text-2xl text-black-500 underline decoration-[0.18rem] underline-offset-[12px] hover:text-gray-500">
-            Directions
-          </h1>
+        <div className="flex cursor-pointer" onClick={() => setIsOpen(!isOpen)}>
+          <h1 className={activeSectionStyling()}>Directions</h1>
         </div>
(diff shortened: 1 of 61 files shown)


pages/studentPortal/web/React/assignments/template.tsx
```diff
@@ -11,44 +11,72 @@ import { Button } from "../../../../../components/ui/Button";
...
+      const currentAssignment = userAssignments.filter(
+        (assignment) => assignment.assignmentId === assignmentId
+      );
+      let assignmentReviewLink = currentAssignment.map(
+        (assignment) => assignment.review_link
+      );
+      let assignmentSubmissionLink = currentAssignment.map(
+        (assignment) => assignment.submission_link
+      );
+      if (assignmentReviewLink != null) {
```
iamnotreddy, line 43: can you just refer to currentAssignment.review_link and currentAssignment.submission_link?

pages/studentPortal/web/React/assignments/template.tsx
```diff
@@ -11,44 +11,72 @@ import { Button } from "../../../../../components/ui/Button";
...
+    useLazyQuery<FetchUserAssignmentSubmissionsDataResponse>(
+      FETCH_USER_ASSIGNMENT_SUBMISSIONS,
+      {
+        variables: {
+          user_id: user.uid,
+          assignmentId: assignmentId,
+        },
+
+        onCompleted: (data: FetchUserAssignmentSubmissionsDataResponse) => {
+          if (data.user_assignment_submissions.length > 0) {
```
iamnotreddy, line 65: i think the logic for checking the DB response and the redux state can be handled by a singular function so you don't have to update logic in both places

you'll just have to write the function so it can work in either case

pages/studentPortal/web/React/assignments/template.tsx
kwiedenhaupt: I can if I change the filter I used on line 34 to find... or do the first record..

I changed to this:
 const currentAssignment = userAssignments.find(
        (assignment) => assignment.assignmentId === assignmentId
      );

and then I can access currentUserAssignment.review_link etc directly.

kwiedenhaupt:
Screen Shot 2023-01-08 at 8 36 16 AM
updated UI

kwiedenhaupt:
image
updated UI

components/coding/studentPortal/AssignmentComponent.tsx
```diff
@@ -8,9 +8,13 @@ export type Hint = {
   icon?: string;
   link?: string;
 };
+
 export enum Stage {
+  // eslint-disable-next-line no-unused-vars
```
iamnotreddy, line 13: i think you can actually just store this in the template and you won't get this error because it's being used there

eslint is here to make us write better code so let's not disable unless we have to

components/coding/studentPortal/AssignmentInputBox.tsx
```diff
@@ -33,6 +34,17 @@ export const AssignmentInputBox: React.FC<AssignmentInputBoxProps> = ({
     }
   );
 
+  const validateInput = () => {
+    if (
+      !submissionInput.includes("codesandbox") ||
+      !submissionInput.includes("https")
+    ) {
+      alert("Please submit a valid submission link.");
+      return false;
```
iamnotreddy, line 43: how about we disable the button if the validateInput function returns false.. could do something like in the goals feature if the text exceeds the number of characters

components/coding/studentPortal/AssignmentInputBox.tsx
```diff
@@ -33,6 +34,17 @@ export const AssignmentInputBox: React.FC<AssignmentInputBoxProps> = ({
     }
   );
 
+  const validateInput = () => {
+    if (
+      !submissionInput.includes("codesandbox") ||
+      !submissionInput.includes("https")
```
iamnotreddy, line 40: i think we can just check whether it includes codesandbox.com

iamnotreddy (approved):

components/coding/studentPortal/AssignmentComponent.tsx
```diff
@@ -30,7 +38,7 @@ export type AssignmentComponentData =
   | {
       component: "output";
       title: string;
-      screenshot: string;
+      screenshotOrVideoId: ScreenshotOrVideoId;
```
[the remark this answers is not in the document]
kwiedenhaupt: good idea.

components/coding/studentPortal/AssignmentComponent.tsx
kwiedenhaupt: I tried that and it still didn't read it lol. said the variable was not being used, but I will move it to template regardless

components/coding/studentPortal/AssignmentInputBox.tsx
kwiedenhaupt: great idea!

PR #852, opened 2023-01-21
Create User Roles and display Coaching Dashboard on Sidebar

kwiedenhaupt:
Changes made to Hasura Tables:
1. Created a role column in the users table with the type of text.  Set default to student.
2. created an enum table called role and inserted two values "coach" and "student".  more could be added later easily.
3. created a foreign key in the users table linking to the enum role table.  This is the only way in hasura that I know of that you can create a drop down list of values that can later be altered as needed. 
4. ran an SQL statement in hasura to check the coaches table for ids and update the users table's row column to coach for those users. 
5. added lucky and i manually to coaches.
image

Changes made to the code:
1. created a fetchUserRole query with types that will return the value of role from the users table for the logged in user
2. created an isRoleCoach redux state on the sidebarSlice that will save the results of that query.
3. implemented the fetchUserRole query on the layout page and saved results to the redux state.
4. recalled the redux state in the sidebar component in order to dispatch the coaching dashboard page if the role was coach only. 
5. changed the title at the top of the sidebar depending on the correct role (isRoleCoach redux boolean).

View for Coaches:
image

View for students:
image

Open to comments regarding the placement and naming of the coaching dashboard.


graphql/fetchUserRole.ts
```diff
@@ -0,0 +1,24 @@
...
+  }
+`;
+
+export type FetchUserRoleData = {
+    users: UserRoleData;
+}
+
+
+export type UserRoleData = {
+    userRole: UserRole;
```
iamnotreddy, line 19: instead of defining a separate UserRole type, pretty sure you can just declare userRole only takes two specific strings "student" or "coach" i.e userRole: "student" | "coach"

kwiedenhaupt: makes sense.  I changed it.  I thought we needed to type out the data for the fetchUserRole query we were implementing on the layout.tsx.

iamnotreddy (commented):
Awesome work on this PR. I like the enum approach you came up with, think it's pretty flexible for potential future cases. Let's get a second look from vithushan

@vithushan19:
1. Can you double-check the DB changes here? Feel like it's pretty aligned with the example in the hasura [docs](https://hasura.io/docs/latest/schema/postgres/enums/). Side note - Kari and I were discussing some cool ways to use the role down the road, like having a "prospect" or "free-trial" role which gates specific content
2. How do you feel about the positioning and / or naming of the "Coaches Dashboard" on the Sidebar?

graphql/fetchUserRole.ts
```diff
@@ -0,0 +1,24 @@
+import { gql } from "@apollo/client";
+
+export const FETCH_USER_ROLE = gql`
+query FetchUserRole($_id: String = "") {
```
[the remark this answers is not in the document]
kwiedenhaupt: I changed both the fetchUserRole query and query implementation on the layout.tsx page.  It was a mistake.  It works, but is probably easier read with simply "id".  I'm calling that query to fetch from the users table directly and that's why it's not userId, or user_id, etc.  The column in users is just simply id.

redux/sidebarSlice.ts
```diff
@@ -5,6 +5,7 @@ export type SidebarProps = {};
 
 export type SidebarPage =
   | "dashboard"
+  | "coaching dashboard"
```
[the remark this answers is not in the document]
kwiedenhaupt: good idea!

vithushan19:
this looks great to me! nice work @kwiedenhaupt Very cool to see how you are expanding on your hasura knowledge to support enums. You're starting to show more flashes of not just applying patterns that you've seen before, but extending patterns to problems that you've never seen. Great work, I'm also excited to start using this in my work to help track all the students. 

I'm good with keeping the coaches dashboard in the sidebar as long as it's hidden to most students.

components/coding/studentPortal/Layout.tsx
```diff
@@ -33,6 +37,20 @@ export const Layout: React.FC = ({ children }) => {
     fetchPolicy: "cache-and-network",
   });
 
+  const {} = useQuery<FetchUserRoleData>(FETCH_USER_ROLE, {
```
vithushan19, line 40: I might move this query into the sidebar, so that the query is happening closer to the component that needs to use it.

vithushan19 (approved):

components/coding/studentPortal/Sidebar.tsx
```diff
@@ -81,6 +81,7 @@ export const Sidebar: React.FC = () => {
           {user && (
             <div className="w-full">
               <p className="w-full ml-4 font-bold">{user.displayName}</p>
+              <p className="ml-4 font-medium">Student</p>
```
iamnotreddy, line 84: what if we just display the role from the redux slice here so it shows either "student" or "coach"? maybe just use a string method to make it upper case

iamnotreddy (approved):
🚀

components/coding/studentPortal/Layout.tsx
kwiedenhaupt: perfect.  did that!

components/coding/studentPortal/Sidebar.tsx
kwiedenhaupt: great!
I had set the redux to isRoleCoach and just realized that you expected it to be a string, so I changed the redux to save the actual string "student" or "coach" and then did the .toUpperCase()

redux/profileSlice.ts
```diff
@@ -6,7 +6,7 @@ export type ProfileState = {
   userProfileData: UserProfileData;
   userBadgeCount: number;
   totalBadgeCount: number;
-  isRoleCoach: boolean;
+  userRole: string;
```
iamnotreddy, line 9: "student" | "coach"

PR #854, opened 2023-01-23
Save Badges Button -Achievement Component

LuckyHariharan:
In this PR I've implemented a save button so signed in users, and coaches can adjust and save their badge selections.
https://www.loom.com/share/fa7fd784eb144337a5a5ebde0e67f084

Key Notes:
- The delete mutation does a call for each badge individually, as I was having trouble passing through all the variables and processing it in one call. 
- The insert mutation runs only once, by passing the array addedBadges as an $object 
- I spent some time re-organizing and re-naming functions, variables and types for clarity.

Hey @raveenreddy0 and @vithushan19, I would appreciate any and all feedback on this PR!


components/coding/profileV2/AchievementComponent.tsx
```diff
@@ -1,172 +1,161 @@
...
-              return {
-                ...badge,
-                isAwarded: !badge.isAwarded,
-              };
-            } else {
-              return badge;
-            }
-          }),
-        };
-      });
```
LuckyHariharan, line 49: Removed Transform function and instead used flatMap to mutate the nested list of badges into a single array of badges

components/coding/profileV2/AchievementComponent.tsx
```diff
@@ -1,172 +1,161 @@
...
+          .user_coding_badges.length === 0
+      ) {
+        addedBadgesTemp.push(badge);
+      } else if (
+        badge.user_coding_badges.length === 0 &&
+        initialBadgeList[index].user_coding_badges.length > 0
+      ) {
+        removedBadgesTemp.push(badge);
+      }
+    });
```
LuckyHariharan, line 68: This function takes the changedBadgesList and initialBadgeList from findBadgeDiff and creates a list of addedBadges and removedBadges to send to the graphql mutation

components/coding/profileV2/AchievementComponent.tsx
```diff
@@ -1,172 +1,161 @@
...
+      );
+      return !initialBadge;
+    });
+    //
+
+    return {
+      initialBadgeArray,
+      finalBadgeArray,
+      changedBadgesList,
+    };
```
LuckyHariharan, line 106: findBadgeDiff takes the difference between the queried badges (initial) and updated badges by the student/coach (finalbadge) and creates a new list containing only the badges that have changes.

graphql/coding/userBadges/updateUserCodingBadges.ts
```diff
@@ -0,0 +1,31 @@
+import { gql } from "@apollo/client";
+export type user_coding_badges_insert_input = {
```
LuckyHariharan, line 2: For some reason this type has to be this name otherwise I run into an error that says something like Type_name_selected is not user_coding_badges_insert_input.

components/coding/profileV2/AchievementComponent.tsx
```diff
@@ -1,172 +1,161 @@
...
-  id: number;
-  title: string;
-  userCodingBadge: userCodingBadge[];
-  isAwarded: boolean;
-};
-export type unit = unitProps[];
-const AcheivementComponent = ({ user, isEditable = false }) => {
-  const { data } = useQuery(FETCH_CODING_BADGES, {
+
+const AcheivementComponent = ({
```
iamnotreddy, line 23: let's get a handle on the naming inconsistency here.. I see the file name is AchievementComponent but you're declaring and exporting the functional component as "AcheivementComponent"... this is getting pretty confusing and is gonna lead to errors down the road

graphql/coding/userBadges/fetchUserBadges.ts
```diff
@@ -17,14 +17,18 @@ export const FETCH_CODING_BADGES = gql`
   }
 `;
 
-type Response = {
+export type Response = {
```
iamnotreddy, line 20: specific need to export this type globally? I could see this causing some issues because it's named pretty generically

graphql/coding/userBadges/updateUserCodingBadges.ts
iamnotreddy: "user_coding_badges_insert_input" shouldn't be a type - it's part of the gql syntax for bulk inserting several values into a DB

I think you're getting a bit confused here - take a look at how skill ratings feature works and how multiple skill ratings are updated into the DB

pages/profile/v2/index.tsx
```diff
@@ -60,7 +60,7 @@ export default function Profile() {
       </div>
       <div className="grid">
         <ExpandableContainer open={true} title={"Achievements"}>
-          <AchievementComponent user={user} />
+          <AchievementComponent userId={user.uid} isEditable={false} />
```
LuckyHariharan, line 63: set isEditable to be false so no one can edit badges. Next step is to set isEditable to be true if the logged in user is that user or a coach.

components/coding/profileV2/AchievementComponent.tsx
```diff
@@ -1,172 +1,161 @@
...
 
-      return updatedData;
+  const [saveAddedBadges] = useMutation(INSERT_USER_CODING_BADGES, {
+    refetchQueries: [{ query: FETCH_CODING_BADGES }],
+  });
+  const [saveRemovedBadges] = useMutation(DELETE_USER_CODING_BADGES, {
+    refetchQueries: [{ query: FETCH_CODING_BADGES }],
+  });
+
+  const separateBadges = (
```
iamnotreddy, line 46: nit: it's not 100% apparent by the name that this is a function

components/coding/profileV2/achievementcomponenets/CodingBadgeUnit.tsx
```diff
@@ -0,0 +1,68 @@
+import React from "react";
```
iamnotreddy, line 1: let's use snakeCase and correct the component spelling for the folder name

I'd probably also put the main AchievementComponent file in here if we're going to have a separate folder for this set of components

LuckyHariharan:
> We're going to need some more work on this PR. A few things:
> 
> * I'm not actually able to see the edit functionality when I navigated to /profile/v2... can you confirm that's working?
> * Let's make sure the naming for "Achievement" is consistent across file name, components, function names etc... we need to be consistent
> * I don't fully understand what you're doing with the findBadgeDiff, separateBadges, and the handleOnSaveButton click. I think you only need two of these. Let me give you an idea below
> * I think the findBadgeDiff function can return two object arrays: the badges to add and the badges to delete, which can be passed into the respective add / delete mutations. Inside a handler function, you can call the findBadgeDiff function, then pass the appropriate object arrays into the appropriate mutation functions.
> * re: bulk deleting, have you taken a look at this hasura [doc](https://hasura.io/blog/bulk-update-and-delete-mutations-in-graphql-6568f47ad0b9/)
> 
> Left a couple other comments below. I think you're getting a bit confused with how bulk mutations are working - probably a good idea to review some examples of these inside the Skillify codebase (skill ratings, goals, etc.). Happy to do some more workshopping later this week

Was able to address all of you feedback except for the bulk deletion. Thanks for linking the documentation, but still having trouble with it.

components/coding/profileV2/achievement_components/AchievementComponent.tsx
```diff
@@ -0,0 +1,120 @@
...
+  const [saveAddedBadges] = useMutation(INSERT_USER_CODING_BADGES, {
+    refetchQueries: [{ query: FETCH_CODING_BADGES }],
+  });
+  const [saveRemovedBadges] = useMutation(DELETE_USER_CODING_BADGES, {
+    refetchQueries: [{ query: FETCH_CODING_BADGES }],
+  });
+  // move handler inside AC component, and add type props
+  // keep findBadgeDiff
+  const handleOnSaveButtonClick = () => {
+    const addedBadges = findBadgeDiff(
```
iamnotreddy, line 46: this function is so much cleaner 🙏

components/coding/profileV2/achievement_components/AchievementComponent.tsx
```diff
@@ -0,0 +1,120 @@
...
+    ).removedBadges;
+
+    if (addedBadges.length > 0) {
+      saveAddedBadges({
+        variables: {
+          objects: addedBadges,
+        },
+      });
+    }
+    if (removedBadges.length > 0) {
```
iamnotreddy, line 64: this is obviously not the most ideal since we can't bulk delete a list of objects, but i love that you came up with this workaround for deleting... might be the next best thing we can do!

components/coding/profileV2/achievement_components/CodingBadgeUnit.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { IntroCourseUnit } from "../../../../graphql/coding/userBadges/fetchUserBadges";
+
+import { CodingBadge } from "../../../../graphql/coding/userBadges/fetchUserBadges";
+
+type CodingBadgeUnitProps = {
+  disabled: boolean;
+  badge: CodingBadge;
+  setUnitBadges: React.Dispatch<React.SetStateAction<IntroCourseUnit[]>>;
+};
+// maybe use IndividualCodingBadge
```
iamnotreddy, line 11: remove this

components/coding/profileV2/achievement_components/CodingBadgeUnit.tsx
```diff
@@ -0,0 +1,68 @@
+import React from "react";
```
iamnotreddy, line 1: can we change the folder name from snake_case please thank you 🤣

components/coding/profileV2/achievement_components/findBadgeDiff.ts
```diff
@@ -0,0 +1,54 @@
+import { IntroCourseUnit } from "../../../../graphql/coding/userBadges/fetchUserBadges";
+import { CodingBadge } from "../../../../graphql/coding/userBadges/fetchUserBadges";
+
+const findBadgeDiff = (
```
iamnotreddy, line 4: this function has come a looooooong way from when we first started... great work!

graphql/coding/userBadges/updateUserCodingBadges.ts
```diff
@@ -0,0 +1,27 @@
+import { gql } from "@apollo/client";
+
+export const DELETE_USER_CODING_BADGES = gql`
```
iamnotreddy, line 3: 😭

iamnotreddy (approved):
just want to reiterate - really great job pushing this through past the finish line, i know it wasn't easy... non-trivial technical decision, couple curve balls etc.

you'll be better for going through this PR!

PR #855, opened 2023-01-28
Edit Flow Achievement Component

LuckyHariharan:
In this PR I've:

- Created a separate component to enable coaches to edit a users badges.
- Implemeneted Promise.all to prevent unique_id_constraint errors if users submit several badge changes in a session 
image


components/coding/profileV2/achievement_components/AchievementComponent.tsx
```diff
@@ -54,31 +57,39 @@ const AchievementComponent = ({
...
-        saveRemovedBadges({
-          variables: {
-            badgeId: badge.badgeId,
-            userId: badge.userId,
-          },
-        });
-      });
-    }
-    alert("Your badge selections have been updated.");
+    Promise.all([
```
vithushan19, line 60: Very cool to see you using promises in such a complex way, awesome job. Lets chat about promises this week and I'll assess your knowledge of them. If you understand promises well, I'll create an advanced JS badge for you on skillify and issue it to you.

components/coding/profileV2/achievement_components/AchievementComponent.tsx
```diff
@@ -113,6 +124,22 @@ const AchievementComponent = ({
...
+      <div className="flex place-content-between space-x-4 px-4">
+        <Button
+          label={"Save"}
+          onClick={() => handleOnSaveButtonClick()}
+        ></Button>
+        <button
+          onClick={() => setEditMode(!editMode)}
+          className="w-5 h-5 cursor-pointer hover:text-yellow-600"
+        >
+          {editMode ? (
```
vithushan19, line 136: instead of using the ternary this way I would do something like this

`<PencilAltIcon className={'w-5 h-5 cursor-pointer ${editMode ? "text-yellow-600" : "hover:text-yellow-600"} '} />`

use backticks instead of single quotes though. Thing that would be cleaner.

vithushan19 (commented):
I wonder if we should create two Achievement Components.

`AchievementComponent` and `EditableAchievements`

One should be used on the coaching dashboard and the other should be used on the regular user profile.

We already check that the user is a coach before letting them navigate to the coaching dashboard. 

On the read-only achievement component, when a user clicks on a badge, I want them to navigate to a new page that displays details about a specific badge. Students will be able to save custom notes about each earned badge on this page.

I think your current implementation might get confusing if you need to handle button clicks for the editable version and then use a Link component for the read only.

components/coding/profileV2/achievement_components/AchievementComponent.tsx
LuckyHariharan: Will do! thank you for the feedback

components/coding/profileV2/achievement_components/AchievementComponent.tsx
```diff
@@ -13,18 +13,21 @@ import {
...
-const AchievementComponent = ({
-  userId,
-  isEditable,
-}: AchievementComponentProps) => {
+const AchievementComponent = ({ userId }: AchievementComponentProps) => {
   const [unitBadges, setUnitBadges] = useState<IntroCourseUnit[]>();
+  // editMode set to true only when a coach accesses the page
   const [editMode, setEditMode] = useState(false);
+  // isEditable set to true only when a coach clicks the PencilAltIcon Component
+  const [isEditable, setIsEditable] = useState(false);
```
vithushan-sound, line 30: I might rename this to say `isEditButtonVisible` so the edit button is only visible for coaches. And coaches have to click the edit button in order to turn on edit mode. 

Eventually, a coach should be able to navigate to a badge detail page by clicking on a badge when editMode is off.

vithushan-sound (approved):
this looks good to me! nice work

components/coding/profileV2/achievement_components/AchievementComponent.tsx
vithushan19, line 30: I might rename this to say isEditButtonVisible so the edit button is only visible for coaches. And coaches have to click the edit button in order to turn on edit mode.

Eventually, a coach should be able to navigate to a badge detail page by clicking on a badge when editMode is off.

vithushan19 (approved):
lgtm nice work

components/coding/profileV2/achievement_components/AchievementComponent.tsx
LuckyHariharan: fixed, ty

PR #856, opened 2023-01-28
totalBadgeCount and userBadgeCount added to coaching dashboard

kwiedenhaupt:
dashboard view
image
profileV2 view
image
Note @vithushan19 regarding redux --
1. did a lot of testing for totalBadgeCount regarding redux.  I ran my FETCH_TOTAL_USER_BADGES_COUNT query it in the ProfileHeaderComponent and dispatched it to a redux state.  But now I need to pull that number from redux and use it in ProfileDetailCard component.  when the query is in ProfileHeaderComponent, the redux does not persist between pages. 
 It is only available on the profile page. when I navigate to the coaching dashboard, it always resets to 0, which is the initial.  If I put the query in the Sidebar, that is always loaded no matter what page, the redux state is available between pages and components at the correct value.  Any idea what is happening or why it's happening?
 
2.  Calculated and deployed the userBadgeCount for only enrolled users -- 
@vithushan19 , is it correct that the earned_badges field in the users table that is used to set the ProfileDetailCard is not dynamic? none of the students who are enrolled have a number in that column, and all but two have earned badges.

3.  Therefore, I had to do some manipulating of data. First I ran a new query named FETCH_EARNED_BADGES that fetched a list of earned badges for all users.  I then filtered from the data response of previous query on the page, FETCH_USER_PROFILE_CARD, to only get a list of userIds of enrolled users.  I named this list enrolledUsers and used that list as a variable in the query FETCH_EARNED_BADGES. This gave me a list of earnedBadges for all enrolled users.  I needed data from the user_coding_badges and users tables both and they are not connected by foreign key, and that is why I chose to do it that way.

4. I then manipulated the list of all badges for enrolled users to aggregate to a total for each user and then saved the object of key:value pairs of userId:userBadgeCount to a redux state called earnedBadges.  It could be used to write back to the database for badges_earned, I guess if necessary.

5. I then added a useEffect to re-dispatch the userList so that it could be transformed with the badgesEarned data from the query along with the data from the previous query, which was saved into userList.  Essentially this transformer just adds the  userBadgeCount into the badges_earned field that was a dummy field and not current.  

6.  I then dispatched the data from the transformer into a final userList redux state which Lucky had previously used to map the ProfileDetailCard.  

7. implemented the totalBadgeCount reducer on ProfileDetailCard so that it shows up as the correct number rather than the hard-coded 50.

SIDENOTES:
1. Created a coachingDashboard file in the API folder to make the CoachingDashboard component less bulky.
2.  updated totalBadgeCount query to be distinct on badge -- it was catching too many before!
3.  updated fetchEarnedBadges query to be in ascending order by userId because it helped me in matching up the list in the transform function so that I could easily map the userBadgeCounts to the ProfileDetailCards


vithushan19:
1. Im not sure why it's not persisting, but you're making an assumption about the order that these components are loaded. I wouldn't even put that number in the redux state. I would just fetch the data from hasura in both components.

We could use next js server side props to prefetch that number ahead of time.

vithushan19:
2. Yeah that column shouldn't be used, that was me testing something out. We should delete it so future team members aren't confused. It's not dynamically calculated or even updated. So better to calculate that number a different way by querying the user coding badges table and filtering for that users id

vithushan19:
3. I think that's a great approach, that's how I would have done it. Glad you thought of it on your own! Really happy to see that.

vithushan19:
5. I need to see how you did this. Ideally all this calculations happen when the component is rendered in the use effect and doesn't need to be retriggered. This may be reasonable but we should talk about it.

components/coding/studentPortal/ProfileDetailCard.tsx
```diff
@@ -16,12 +19,13 @@ type ProfileDetailCard = {
 function ProfileDetailCard({
   avatar,
   name,
-  link,
   joinDate,
   badges,
   currentBadge,
   nextGoal,
 }: ProfileDetailCard) {
+  const { totalBadgeCount } = useSelector(profileSelector);
```
vithushan19, line 27: We should pass this as a prop like the other props here like currentBadge, name, etc...

Let's not use the selector here.

components/coding/studentPortal/Sidebar.tsx
```diff
@@ -76,6 +79,19 @@ export const Sidebar: React.FC = () => {
 
   const router = useRouter();
 
+  const { loading: totalUserBadgeCountLoading } =
+    useQuery<FetchUserBadgesCountResponse>(FETCH_TOTAL_USER_BADGES_COUNT, {
```
vithushan19, line 83: I don't want the sidebar to become bloated and become the go-to place to query all sorts of data and update redux.

Let's figure out somewhere else to put this query. Can we put this on the profile page or coaching dashboard page maybe? 

Or can we adapt the initial query that fetches all the enrolled students to calculate total badge as well?

Happy to chat about it

graphql/fetchUserProfileCard.ts
```diff
@@ -20,19 +20,19 @@ export type FetchUserProfileCardResponse = {
...
+query fetchUsers {
+  users(where: {enrolled: {_eq: true}}, order_by: {id: asc}) {
+    id
+    name
+    link
+    profile_image
+    badges_earned
+    created_at
+    current_badge
+    coding_badge {
```
vithushan19, line 32: Can we just use the length of this array to determine the number of earned badges instead of making a new query?

pages/coaching-dashboard.tsx
```diff
@@ -1,24 +1,56 @@
...
+    variables: {
+      enrolledIds: enrolledUsers,
+    },
+    onCompleted: (data) => {
+      dispatch(setEarnedBadges(data.user_coding_badges));
+    },
+  });
+
+  useEffect(() => {
+    const aggregatedBadgeCount = earnedBadges.reduce((acc, badgeId) => {
```
vithushan19, line 41: This is cool that you figured out how to transform the data, but it's confusing for new developers. I think we can figure out an easier way.

pages/coaching-dashboard.tsx
```diff
@@ -27,25 +59,29 @@ const coachingDashboard = () => {
...
-                currentBadge={it.coding_badge}
-                nextGoal={""}
-                link={it.link}
-              />
+        {userList.map((it, index) => {
+          return (
+            <div key={index}>
+              <Link href={"profile/" + it.link}>
+                <div className="">
+                  <ProfileDetailCard
```
vithushan19, line 67: Could just feed in earned badges as a prop here instead of storing in redux.

pages/skillRatings.tsx
```diff
@@ -71,6 +73,7 @@ export default function SkillRatings(props) {
 
   return (
     <div className="flex flex-row overflow-auto-bg-scroll">
+      {totalBadgeCount}
```
vithushan19, line 76: Intentional?

redux/userSlice.ts
```diff
@@ -1,26 +1,34 @@
 import { createSlice, PayloadAction, Slice } from "@reduxjs/toolkit";
+import { EarnedBadges } from "../graphql/fetchEarnedBadges";
 
 import { Users } from "../graphql/fetchUserProfileCard";
 import { RootState } from "./rootReducer";
 
 export type userState = {
   userList: Users[];
```
vithushan19, line 8: I wonder if this slice is even needed. I know lucky created it.

vithushan19: But I might ask him to remove it

components/coding/studentPortal/ProfileDetailCard.tsx
kwiedenhaupt: fed it in as a prop. thanks!

graphql/fetchUserProfileCard.ts
kwiedenhaupt: No, unfortuantely, that number is not accurate with the current foreign key.  I did keep the earnedBadges list as explained above, for now, but would like to look into implementing a foreign key.

pages/coaching-dashboard.tsx
kwiedenhaupt: great idea.  I removed it.  thank you!

redux/userSlice.ts
kwiedenhaupt: I removed it and just used local state.

components/coding/studentPortal/Sidebar.tsx
iamnotreddy: @kwiedenhaupt - try making userId in user_coding_badges foreign key to the users table and id column. I think you should be able to create an array relationship between users and the user_coding_badges table, at which point you could adapt the initial query that fetches enrolled students like Vithushan mentioned

I feel like this should work but not 100% sure. Try giving it a shot

components/coding/studentPortal/ProfileDetailCard.tsx
```diff
@@ -40,7 +44,7 @@ function ProfileDetailCard({
         <BadgesDisplayedComponent
           earnedBadges={badges}
```
iamnotreddy, line 45: if a user hasn't earned any badges, the UI shows: 
Screen Shot 2023-01-30 at 10 49 22 AM

need to add some logic to show 0 in this case

graphql/fetchTotalUserBadgesCount.ts
```diff
@@ -2,10 +2,10 @@ import { gql } from "@apollo/client";
 
 export const FETCH_TOTAL_USER_BADGES_COUNT = gql`
```
iamnotreddy, line 3: i think you could actually just use the `coding_badges_aggregate` endpoint to get the total available badges

right now you're querying from the user-level table and using distinct to get the unique number of badgeIds that exist.. this isn't wrong but using distinct is a more expensive operation.. in our case there probably isn't a noticeable performance hit but imagine if we had 100s of thousands of badges or something.

instead you could just count the total number of rows in the coding_badges table... the primary key in this table is just a single badge, so something like this query should do the same thing:

```
query MyQuery {
  coding_badges_aggregate {
    aggregate {
      count
    }
  }
}
```

graphql/fetchUserProfileCard.ts
```diff
@@ -1,38 +1,50 @@
...
   coding_badge: CodingBadge;
+  user_coding_badges_aggregate: BadgesEarnedCount;
 };
 
 export type CodingBadge = {
   title: string;
   image: string;
 };
 
+export type BadgesEarnedCount = {
```
iamnotreddy, line 19: cc: @LuckyHariharan 

I learned a new typescript trick recently - when we have nested fields, we don't _have_ to declare a new type for each nested field. Take a look at this:

```
export type Users = {
  id: string;
  name: string;
  link: string;
  profile_image: string;
  created_at: string;
  current_badge: number;
  coding_badge: CodingBadge;
  user_coding_badges_aggregate: {
    aggregate: {
      count: number;
    };
  };
};
```
you don't need the BadgesEarnedCount and BadgesEarnedNested types any more

In some cases it does make sense to have separate nested types like coding badges

iamnotreddy (changes requested):

pages/coaching-dashboard.tsx
```diff
@@ -8,17 +8,36 @@ import {
...
 
   const { userList } = useSelector(userSelector);
+  const { totalBadgeCount } = useSelector(profileSelector);
 
   const { loading, data } = useQuery<FetchUserProfileCardResponse>(
     FETCH_USER_PROFILE_CARD,
     {
-      onCompleted: () => dispatch(setUserList(data.users)),
+      onCompleted: () =>
+        data.users.length > 0 ? dispatch(setUserList(data.users)) : loading,
```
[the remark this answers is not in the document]
kwiedenhaupt: great idea

graphql/fetchTotalUserBadgesCount.ts
kwiedenhaupt: thank you for pointing that out.  that makes a lot of sense in terms of efficiency. I really appreciate thinking this way all the time!

graphql/fetchUserProfileCard.ts
kwiedenhaupt: Great idea.  I had a couple of those on this PR and I changed them all!

pages/skillRatings.tsx
```diff
@@ -21,6 +22,7 @@ export default function SkillRatings(props) {
   const dispatch = useDispatch();
   const { skillRatings } = useSelector(skillRatingsSelector);
   const { user } = useAuth();
+  const { totalBadgeCount } = useSelector(profileSelector);
```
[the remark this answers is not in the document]
kwiedenhaupt: no.  i was checking to see if the redux would store between pages or if it was just the coaching-dashboard.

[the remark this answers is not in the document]
kwiedenhaupt: removed! as it was a mistake.

iamnotreddy (approved):
nice work!

PR #857, opened 2023-01-30
Changes to styling of the Coaching Dashboard Cards

kwiedenhaupt:
image
A couple thoughts:
1. For now I have used the enrolledUsers list I pulled from the initial query ran on the coaching-dashboard from FETCH_USER_PROFILE_CARD and fed in the list of enrolledUsers into my query for the goals and manipulated the data to display from there.  We could think about linking the tables together as a foreign key -- from the user_goals_table and linking the enrolled users column to that table directly.

2.  I couldn't figure out how to get a div circle that remained w-16 and h-16 no matter what the length of the text was for the goal....the longer goal text would cause the circle to squish.  That doesn't happen if you use an image, but we wanted to put the date inside the circle.  

3.  I couldn't figure out how to fix the date bug where if there is no date because there is no goal it changes it automatically to 12/31 when it formats.

4.  Let me know what you think about the header styling -- if you want them up in the same line like that or not.


iamnotreddy:
I think cool with the header style, it's a more efficient use of space. But we need to work on the alignment. I think the name, badge count, and join date need to vertically aligned with the avatar. I think you're missing an items-center or justify-center in the tailwind class. Same goes for the bottom half of the card. The actual focus text should be vertically aligned with the circle.

Instead of using absolute pixel sizes, can we try using relative sizes?

The cards are different sizes in some cases so we should try to fix that. I'm thinking the ProfileDetailCards should be part of a grid so their sizes are fixed. I think the cards themselves can be inside a flex container.

re: the date bug, I'm not sure I follow. Can you give me some more details?

kwiedenhaupt:
image
This is the updated styling of the cards.  
I also pulled the data manipulation logic into a useEffect, because the data was not loading into goalsList before the return sometimes and therefore the goals fields failed to populate on occasion.  it seems to work now.  I also changed the layout to a 4 column grid, and that seemed to take care of the alignment issues, as well as the standard circle around the goal date.  

Relative pixel sizes for the images and the circle div do not work, as the height does not match the width and I can't seem to get a circle.  I think the 20px and 16px look good though.

Card sizes are standard.

you can let me know what you think of the underlining of the text at the bottom??? I'm not sure.

pages/coaching-dashboard.tsx
```diff
@@ -30,6 +38,40 @@ const coachingDashboard = () => {
       },
     }
   );
+  const enrolledUsers = userList.map((user) => user.id);
```
iamnotreddy, line 41: you're not actually using this - you define enrolledUsers again in the useEffect where it's actually used

pages/coaching-dashboard.tsx
```diff
@@ -48,13 +90,14 @@ const coachingDashboard = () => {
   return (
     <div className="flex flex-col p-4 m-4 ">
       <p className="mb-8 text-3xl font-bold">Coaching Dashboard</p>
-      <h2 className="mb-4">Enrolled Students</h2>
       <div className="grid grid-cols-1 gap-4 sm:grid-cols-3">
         {userList.map((it, index) => {
+          const completionDate = goalCompletionDateList[index];
+          const completedGoal = completedGoalsList[index];
           return (
             <div key={index}>
```
iamnotreddy, line 98: you can actually get rid of this div and put the same key prop inside the Link component right below

components/coding/studentPortal/ProfileDetailCard.tsx
```diff
@@ -21,34 +22,55 @@ function ProfileDetailCard({
...
-  nextGoal,
+  completedGoal,
+  completedDate,
 }: ProfileDetailCard) {
   return (
-    <div className="flex flex-col h-full gap-4 text-sm border-2 bg-slate-50 text-slate-800 border-slate-800 hover:bg-violet-100">
-      <div className="flex flex-col items-center">
+    <div
+      className="grid grid-cols-4 h-full w-full text-sm border-2 bg-slate-50 text-slate-800 border-slate-800 hover:bg-violet-100"
+      style={{ height: "340px", objectFit: "contain" }}
```
iamnotreddy, line 31: I took this line out and everything looked the same -- are you sure we need this?

trying to stay away from absolute sizing as much as possible since that can cause issues for responsiveness on different screen sizes

components/coding/studentPortal/ProfileDetailCard.tsx
```diff
@@ -21,34 +22,55 @@ function ProfileDetailCard({
...
+      <div className="ml-4 col-span-4 p-1 mb-4">
+        <p className="ml-2 font-bold">Current Focus:</p>
         <div className="flex items-center p-2 ">
-          <img src={currentBadge?.image} className="w-16 mr-4 rounded-full" />
-          <p className="underline "> {currentBadge?.title}</p>
-        </div>{" "}
-        <p className="underline">Next Goal: {nextGoal}</p>
+          <img
+            src={currentBadge?.image}
+            className="col-span-1 items-center w-16 mr-5 rounded-full"
```
iamnotreddy, line 58: i got relative sizing to work, you could try w-1/6 here instead of w-6

components/coding/studentPortal/ProfileDetailCard.tsx
```diff
@@ -21,34 +22,55 @@ function ProfileDetailCard({
...
-          createdAt={joinDate}
-        />
-        <BadgesDisplayedComponent
-          earnedBadges={badges}
-          textSize={"small"}
-          totalBadges={totalBadgeCount}
-        />
-        <p className="font-bold">Current Focus:</p>
+      <div className="ml-4 col-span-4 p-1 mb-4">
+        <p className="ml-2 font-bold">Current Focus:</p>
```
iamnotreddy, line 54: let's get rid of the ":"

components/coding/studentPortal/ProfileDetailCard.tsx
```diff
@@ -21,34 +22,55 @@ function ProfileDetailCard({
...
         <div className="flex items-center p-2 ">
-          <img src={currentBadge?.image} className="w-16 mr-4 rounded-full" />
-          <p className="underline "> {currentBadge?.title}</p>
-        </div>{" "}
-        <p className="underline">Next Goal: {nextGoal}</p>
+          <img
+            src={currentBadge?.image}
+            className="col-span-1 items-center w-16 mr-5 rounded-full"
+          />
+          <p className="col-span-3 underline"> {currentBadge?.title}</p>
```
iamnotreddy, line 60: let's get rid of the underlines also

components/coding/studentPortal/ProfileDetailCard.tsx
```diff
@@ -21,34 +22,55 @@ function ProfileDetailCard({
...
+        </div>
+        <p className="ml-2 font-bold">Last Goal:</p>
+        <div className="flex items-center p-2">
+          <div className="col-span-1 mr-5">
```
iamnotreddy, line 64: col-span is a utility for grids only: https://tailwindcss.com/docs/grid-column

this is inside a flex container so col-span isn't doing anything

components/coding/studentPortal/ProfileDetailCard.tsx
```diff
@@ -21,34 +22,55 @@ function ProfileDetailCard({
...
-    <div className="flex flex-col h-full gap-4 text-sm border-2 bg-slate-50 text-slate-800 border-slate-800 hover:bg-violet-100">
-      <div className="flex flex-col items-center">
+    <div
+      className="grid grid-cols-4 h-full w-full text-sm border-2 bg-slate-50 text-slate-800 border-slate-800 hover:bg-violet-100"
+      style={{ height: "340px", objectFit: "contain" }}
+    >
+      <div className="flex ml-3 place-items-center col-span-4 mt-4">
         <img
-          className="items-center w-20 mt-4 border-3 rounded-full border-slate-800 "
+          className="col-span-1 w-20 my-3 border-3 rounded-full border-slate-800 "
```
iamnotreddy, line 35: another case of relative sizing vs. absolute sizing again

i'd prefer we use something like w-1/4 here, and then a smaller width for the other "circles"

you can always check the tailwind [docs](https://tailwindcss.com/docs/width) to see what width fractions are valid

components/coding/studentPortal/ProfileDetailCard.tsx
```diff
@@ -21,34 +22,55 @@ function ProfileDetailCard({
...
+          <img
+            src={currentBadge?.image}
+            className="col-span-1 items-center w-16 mr-5 rounded-full"
+          />
+          <p className="col-span-3 underline"> {currentBadge?.title}</p>
+        </div>
+        <p className="ml-2 font-bold">Last Goal:</p>
+        <div className="flex items-center p-2">
+          <div className="col-span-1 mr-5">
+            <div className="col-span-1 bg-slate-800 w-16 h-16 border-black rounded-full flex items-center justify-center">
```
iamnotreddy, line 65: want to take a stab at creating a circle using an svg? 

what if you have a parent container that has relative sizing i.e a <div className= 'w-1/6'> then put an svg inside the parent container and use one of the object-fit [properties](https://tailwindcss.com/docs/object-fit)

this might be helpful: https://medium.com/@masonk000/using-svgs-with-tailwind-css-84e9e3526b63

iamnotreddy (changes requested):
You did a good job getting all the cards to have the same size, and getting the inside of the cards look generally pretty good

I think you can improve how you're writing your tailwind though. A couple pointers:
- instead of trying to position individual elements with mr-5 ml-2 etc. or add padding to individual elements, try to add padding or margins in the parent div... this will make your tailwind code cleaner and more consistent
- for example, if you want the whole card to have margin, then add that in the parent div instead of spacing individual elements from the border of the card

components/coding/studentPortal/ProfileDetailCard.tsx
kwiedenhaupt: I took it out -- only if we want to ensure if the text goes longer on the goal that the containers will always remain the same height.  I do agree with you, though, to take it out because the circle is big enough it makes the spacing solidly larger than we need so I removed it!

components/coding/studentPortal/ProfileDetailCard.tsx
kwiedenhaupt: you're right.  it does with the images...just not with styled divs -- great suggestion that I try to change that circle div to an svg.  I will work on that!

components/coding/studentPortal/ProfileDetailCard.tsx
vithushan19: be careful using fractional widths, make sure that it looks good on mobile devices as well. or use tailwind mobile responsive prefixes to look good on mobile. ie `sm:`  or `md:` or `lg:`

vithushan19:
agree with Raveen. In general, avoid putting padding/margin on these reusable components. Much better to have no spacing on the components itself, and just put spacing on the parent containers.

kwiedenhaupt:
I tried to removed unused tailwind code and put all the margin and spacing in the parent containers.  It seems to be a lot cleaner and simpler.

Updated Mobile View
image
updated regular view
image

I think I did get the SVG text to be crisper by adjusting the stroke width to a fraction, changing the text to Arial, the stroke width to an even number seeing as my view box were even coordinates, and changing the textRendering to auto.  I'm not sure that the font weight is of any consequence...
image

iamnotreddy (commented):
lgtm

iamnotreddy (approved):
lgtm

PR #861, opened 2023-02-02
profile routing and structural changes

iamnotreddy:
I've modified the profile flow so the following can happen:

- the v2 profile is now accessed through /profile/v2/[userId] and this component now called the InternalProfile. It can either grab the userId from the URL to fetch the relevant profile information, or it can take a userId as a prop (for the case below) 

- I added a shortcut so you can navigate to /profile/x/[name] i.e /profile/x/Raveen. This is actually a separate component that checks whether the [name] passed is valid against the `link` field in the users table... if it is valid, it grabs the userId and passes that into the InternalProfile component as an optional prop. We should be able to direct an external user (i.e recruiter, hiring manager) to an external version of the profile from here.

- it looks like NextJS title cases anything you pass into the URL. So if we pass in /profile/x/lucky, the router query reads this as "Lucky" which is invalid. Either we figure out how to change the default behavior or we just change everyone's "link" name to TitleCase (the latter is probably easier)

This pattern will allow the current user to look at another user's profile without the current user's data creating conflicts.

I also made some structural changes - the parent profile component now fetches the data and passes it as props into children (ProfileHeader info and badge counts, Goals). Skill Ratings still needs to be modified

What needs to happen next @kwiedenhaupt @LuckyHariharan :

- Modify SkillRatings so it takes in the userId of the profile being viewed as a prop. This way you won't always see your own skill ratings. We need to make sure you can't modify another user's skill ratings.

- the ProfileCards in the coaching dashboard need to link to profile/v2/[userId] now

- the sidebar profile will need to link to profile/v2/[userId] (the current user's userId). Right now it just links to the v1 profile so nothing is broken

- of course we'll remove the whole v2 folder so there's only one profile... might make sense to make the above changes when we're ready for the full cutover


components/coding/profileV2/ProfileHeaderComponent.tsx
```diff
@@ -1,112 +1,42 @@
...
+  totalBadgeCount: number;
 };
 
 export default function UserProfileSection({
-  userId,
+  userProfileData,
+  userBadgeCount,
+  totalBadgeCount,
 }: UserProfileSectionProps) {
-  const dispatch = useDispatch();
```
vithushan19, line 33: wow, this component got much cleaner with your change. nice work.

pages/profile/v2/[userId].tsx
```diff
@@ -0,0 +1,151 @@
...
+export default function InternalProfile({
+  userIdFromLink,
+}: InternalProfileProps) {
+  const { user } = useAuth();
+  const router = useRouter();
+  const dispatch = useDispatch();
+
+  let userId;
+
+  if (userIdFromLink) {
```
Comment (vithushan19) [inline pages/profile/v2/[userId].tsx:49]: nice, maybe a comment here would be useful.

you could also do `const userId = userIdFromLink ? userIdFromLink : router.query.userId`

this way you can have a const variable instead.

pages/profile/x/[userName].tsx
```diff
@@ -0,0 +1,38 @@
...
+        setValidUserId(data.users[0].id);
+      }
+    },
+  });
+
+  if (loading) {
+    return <div>Loading...</div>;
+  }
+
+  if (validUserId) {
```
Comment (vithushan19) [inline pages/profile/x/[userName].tsx:28]: I think this check should be done server side instead of client side. You can move this check to server side by making use of getServerSideProps

https://reacthustle.com/blog/redirect-to-404-nextjs-dynamic-routes
https://nextjs.org/docs/api-reference/data-fetching/get-server-side-props

The react hustle article explains why that approach is better.

vithushan19 (approved):
looks good to me, nice work. left a few comments/questions

PR #864, opened 2023-02-02
Reworking skillRatings for new profile structure

kwiedenhaupt:
1. Moved the query to the [userId] slug page to maintain consistency
2. loaded it in skillRatings redux and drilled it down as a prop to children
3. Placed a check in the query to see if the userId == user.uid (logged in user) -- when true set isEditable
4. drilled isEditable down as a prop to children
5. Changed SkillRowEmoji to disable slider if not logged in user
6. Changed SkillRatingsComponent to remove save button if not logged in user.

When I navigate to x/kari (i.e. logged-in user)
image
When I navigate to x/lucky (i.e. not logged-in user)
image


vithushan19 (approved):
looks good to me!

iamnotreddy (approved):
looks great - nice work!

PR #865, opened 2023-02-02
Set Order for Profile page

LuckyHariharan:
Replacing my previous PR , I swapped the order of the achievements component and skillratings component. The achievements component is now above skillratings.

sorry about that @raveenreddy0


iamnotreddy (changes requested):
i think these divs wrapped around the expandable containers aren't needed but i could be wrong, can you get rid of them?

iamnotreddy (approved):
nice!

PR #866, opened 2023-02-03
Link Coaching Dashboard

LuckyHariharan:
This PR links each card in the coaching dashboard to ...profile/v2/[userId].
Key Notes:
- After linking to profile/v2/[userId], only [userId]'s that were coaches would have editable badges. So I updated AchievementComponent to check if the logged-in user is a coach (not the user whose card is being viewed as it was before)
- After editing and saving badge selections a refresh is required to see the new selections. Simply navigating away and back to the profile/v2/[userId] is not enough to trigger the UI badge update.
image
image


iamnotreddy (approved):
lgtm

PR #869, opened 2023-02-06
it.id instead of it.link 

LuckyHariharan:
(hopefully) final PR for this feature!

Lessgoo


kwiedenhaupt (approved):
Thanks making the change to check for the data when we go between profiles.

PR #870, opened 2023-02-06
adaptation to make spring library work when the containers are set to default closed.

kwiedenhaupt:
The main change is this line of code:
  const [heightRef, height] = useHeight({ on: isOpen });
it sets the function of checking the height of the container to only calculate when the container is open.
Therefore if the container is set to default closed, it will not calculate and set the height at the height of the closed container, but will be only active upon click that opens the container, and will set the dynamic height to the length of the container's contents correctly.

Otherwise the component is set to check the component height upon mount, and it would set it to nothing on default, and the container would only work once open and closed and then opened again.


kwiedenhaupt:
I opened this up in a new branch to ensure that only the spring changes were being brought into main.

PR #872, opened 2023-02-07
Fix bug where the Spring was not working on the expandable container surrounding the Achievements Component

kwiedenhaupt:
1. downloaded the react-use-measure package that contains a useMeasure hook. I think this approach is more responsive to calculating the container height -- especially in the achievements component where the edit button changes the length of the content.
2. Deleted the useHeight custom hook
3. created an additional spring that uses the opacity variable to delay the fading in of the content so that the container opens and then the content fades in after.  

I was having trouble regarding the first open of the container when the containers were all set to closed.  In particular, it was weird because it was happening all the time on mobile and not in dev mode when I went to test out the feature.  

I have tested this solution and think it's closer, but would love feedback regarding what you think concerning if this is smooth enough or how to make it better.


iamnotreddy (approved):
this looks pretty solid to me - really great work! 

i wasn't able to repro any of the other errors we were seeing before. once you merge, it's probably worth checking behavior on mobile to see if you're still running into those issues

PR #873, opened 2023-02-09
Jest Testing blog post

kwiedenhaupt:
1. Wrote the blog post
2. added pics
3. added a function to make the code blocks display decently (CodeBlock)
4. added the post to the index page.
5. side note: added a return for the custom component because there was none.  I tried to use it at one point and got stuck without a render lol

Blog index page
image

Beginning of the post
image

Code snippets in post
image


kwiedenhaupt:
I just realized mobile needs to be fixed -- any suggestions on how best to do that are appreciated :)

kwiedenhaupt:
image
1. Made the post mobile responsive by adjusting the PageLayout -- however, this does change the layout for all the posts. 
2. The only posts that were mobile responsive before were those that used the LessonComponent.
3. the others are now better -- however, they have extra white spacing on the right side that I could fix if we choose to keep some of the adjustments I made.  The videos are better, but the text has extra white spacing around it on the right.
image

Let me know your thoughts on what is best!

vithushan19 (approved):
this looks good to me @kwiedenhaupt 

- can we add a comment to the CodeBlock component and mention that it is only meant to be used within the blog post, and that no one should try to reuse this component elsewhere
- it also looks fine on mobile to me
- your code component looks super complicated just to display some formatted code, but if its working and not affecting other parts of the codebase im ok with this

PR #875, opened 2023-02-10
Launch components assignment / modify general assignment template structure

iamnotreddy:
This PR adds our first assignment to the Skillify React course! I have tested the flow several times to ensure it's fully working. LMK if you see anything that fell through the cracks

Some high-level points:
- the componentsAssignment will serve as the new template
- I more or less completely redesigned the assignment template so it utilizes our Expandable Containers. Lots of stylistic changes to the assignment component library
- I had to make some changes to the staging flow so it accepts a dynamic loom id (before it was hardcoded)
- The assignment is now accessible through the react course page

I don't think this approach is fully scalable yet. If we create several assignments, then change the general structure of the template, we'll have to modify each assignment. In an ideal world, if we change anything on the template, every assignment follows suit automatically. That being said I think this works for now.

Here's what the stages look like: 

Not started: 

image

In review (after submitting a link):

image

Assignment Reviewed: 

image


vithushan19 (approved):
looks good to me, nice work. I agree with your priority to get this out first and we can refactor/templatize assignments going forward.

PR #876, opened 2023-02-12
simplify CodeBlock component and make two other blogs mobile responsive with the iframes

kwiedenhaupt:
1.  The CodeBlock is now much simpler, easier to read, and I better understand the code.  I also customized the theme to be used for the blog post.
2. Modified the Python and Java posts to be mobile responsive, because the iframes 
image
image


vithushan19 (approved):
Nice work! Good job finding the other blog posts and fixing them as well

PR #885, opened 2023-02-20
Radix UI Library Sandbox

kwiedenhaupt:
Here I created/modified and implemented three Radix-UI library components:
1. Dialog
2. Progress
3. Tooltip

I rendered them on a sandbox page 
http://localhost:3001/studentPortal/web/React/radix-ui

but then also on the studentPortal main page.
http://localhost:3001/studentPortal

(DialogComponent is on the index.tsx studentPortal)
(ProgressBarComponent and TooltipComponents are implemented in the Sidebar with case-specific props.

Example of DialogComponent
image

Example of TooltipComponent
image

Example of ProgressBarComponent
image

sandbox page
http://localhost:3001/studentPortal/web/React/radix-ui
image

I created the slidebar just so that you can see how responsive the animation on the sidebar is.  Kind of cool.  that's in a separate component that will not be brought into production.

There is also a corresponding tech spec that is a WIP but can be accessed here:
https://docs.google.com/document/d/13yKOQIJfVnh5lKcae7gkQKW3jlAk8SAuan-vBxjJsao/edit#

At the end of the tech spec contains a general recommendation on the integration of the Radix-UI library in the project.

### DIFF
--- a/components/coding/studentPortal/radix/DialogComponent.tsx
+++ b/components/coding/studentPortal/radix/DialogComponent.tsx
@@ -1,0 +1,46 @@
+import React from "react";
+import { Root, Portal, Content, Overlay, Close } from "@radix-ui/react-dialog";
+import { Button } from "../../../ui/Button";
+import { useRouter } from "next/router";
+import { XIcon } from "@heroicons/react/solid";
+
+const DialogComponent = () => {
+  const router = useRouter();
+
+  const handleContinue = () => {
+    router.push("/studentPortal/web/React/props");
+  };
+
+  return (
+    <Root defaultOpen={true}>
+      <Portal>
+        <Overlay className="bg-gray-300 data-[state=open]:animate-overlayShow fixed inset-0" />
+        <Content className="flex items-center justify-center">
+          <div className="fixed flex flex-col items-center justify-center w-full max-w-md px-3 py-32 mx-auto text-center transform -translate-x-1/2 -translate-y-1/2 bg-white rounded-lg md:max-w-3xl lg:max-w-4xl sm:rounded-3xl left-1/2 top-1/2">
+            <Close asChild>
+              <button className="absolute top-3 right-3 w-6 h-6 flex items-center justify-center rounded-md text-gray-100 duration-500 bg-gray-900 bg-opacity-50 outline-none cursor-pointer hover:bg-opacity-100">
+                <XIcon />
+              </button>
+            </Close>
+            <h2 className="mb-12 text-3xl font-medium font-title md:text-6xl">
+              Wait! Your trial has only{" "}
+              <span className={"text-red-500"}> 5</span> days remaining!
+            </h2>
+            <p className="max-w-md px-2 mx-auto text-xl leading-8 ">
+              Don't forget to submit your first assignment! Time to bring out
+              your inner coder. Get tailored feedback on your first assignment
+              to get a taste of what it's like to be a Skillifyer!
+            </p>
+            <Button onClick={handleContinue} label={"Start Now"}>
+              Join us
+            </Button>
+            <span className="font-mono text-sm font-bold">
+              You'll feel great when you crush it!
+            </span>
+          </div>
+        </Content>
+      </Portal>
+    </Root>
+  );
+};
+export default DialogComponent;
--- a/components/coding/studentPortal/radix/DialogComponent.tsx
+++ b/components/coding/studentPortal/radix/DialogComponent.tsx
@@ -16,20 +16,20 @@ const DialogComponent = () => {
       <Portal>
         <Overlay className="bg-gray-300 data-[state=open]:animate-overlayShow fixed inset-0" />
         <Content className="flex items-center justify-center">
-          <div className="fixed flex flex-col items-center justify-center w-full max-w-md px-3 py-32 mx-auto text-center transform -translate-x-1/2 -translate-y-1/2 bg-white rounded-lg md:max-w-3xl lg:max-w-4xl sm:rounded-3xl left-1/2 top-1/2">
+          <div className="fixed flex flex-col space-y-4 items-center justify-center w-full max-w-md px-3 py-32 mx-auto text-center transform -translate-x-1/2 -translate-y-1/2 bg-white rounded-lg md:max-w-3xl lg:max-w-4xl sm:rounded-3xl left-1/2 top-1/2">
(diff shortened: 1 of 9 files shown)


components/coding/studentPortal/radix/TooltipComponent.tsx
```diff
@@ -0,0 +1,42 @@
...
+  message,
+  active,
+}) => {
+  return active ? (
+    <Provider>
+      <Root>
+        <Trigger asChild>{children}</Trigger>
+        <Portal>
+          <Content
+            className="data-[state=delayed-open]:data-[side=top]:animate-slideDownAndFade data-[state=delayed-open]:data-[side=right]:animate-slideLeftAndFade data-[state=delayed-open]:data-[side=left]:animate-slideRightAndFade data-[state=delayed-open]:data-[side=bottom]:animate-slideUpAndFade text-red-500 select-none rounded-[4px] bg-white px-[15px] py-[10px] text-[15px] leading-none shadow-[hsl(206_22%_7%_/_35%)_0px_10px_38px_-10px,_hsl(206_22%_7%_/_20%)_0px_10px_20px_-15px] will-change-[transform,opacity]"
```
vithushan19, line 27: This is super confusing. What's happening here?

package.json
```diff
@@ -25,6 +25,9 @@
     "@hotjar/browser": "^1.0.6",
     "@magic-ext/oauth": "^2.1.1",
     "@matejmazur/react-katex": "^3.1.3",
+    "@radix-ui/react-dialog": "^1.0.2",
+    "@radix-ui/react-progress": "^1.0.1",
```
vithushan19, line 29: Nice!

pages/studentPortal/web/React/radix-ui.tsx
```diff
@@ -0,0 +1,128 @@
...
+          value={66}
+        >
+          <Indicator
+            className="bg-white w-full h-full transition-transform duration-[660ms] ease-[cubic-bezier(0.65, 0, 0.35, 1)]"
+            style={{ transform: `translateX(-${100 - progress}%)` }}
+          />
+        </Root>
+      </div>
+      <TooltipComponent message="Help me." active={false}>
+        <button className="text-blue-500 shadow-black-500 hover:bg-purple-200 inline-flex h-[35px] w-[35px] items-center justify-center rounded-full bg-white shadow-[0_2px_10px] outline-none focus:shadow-[0_0_0_2px] focus:shadow-black-300">
```
vithushan19, line 58: These tailwind styles are a lot to take in. Can we extract this button into a resuable component and put it in a different file?

tailwind.config.js
```diff
@@ -98,14 +98,47 @@ module.exports = {
...
+        },
       },
       outline: {
         black: "2px solid #000000",
       },
       primary: "#4495f0",
       secondary: "#ff8e4f",
     },
+    animation: {
+      slideDownAndFade: 'slideDownAndFade 400ms cubic-bezier(0.16, 1, 0.3, 1)',
```
vithushan19, line 133: Exciting!

pages/studentPortal/web/React/radix-ui.tsx
kwiedenhaupt: This button I won't use in production... I want to wrap the tooltip component around the sidebar links.

components/coding/studentPortal/radix/TooltipComponent.tsx
kwiedenhaupt: I greatly modified the tailwind classes here.
I chose to have the tooltip always appear from the top and the therefore could eliminate the other three "data-...." for left/right/bottom.  

I also removed the shadow class and simplified it to "shadow-md ring-1 ring-gray-300".  It's not quite as nice, but is similiar.

Finally, to explain..."select-none" doesn't allow cursor clickable events.  leading-none removes any additional vertical space in the tooltip and sets the line height to the text size.  "will-change-[transform,opacity]" signals to the browser that there will be an animation change in opacity/transform will happen here in the tailwind.config.ts file " contentShow: {
          from: { opacity: 0, transform: 'translate(-50%, -48%) scale(0.96)' },
          to: { opacity: 1, transform: 'translate(-50%, -50%) scale(1)' },
        },"

components/coding/studentPortal/radix/DialogComponent.tsx
```diff
@@ -0,0 +1,46 @@
+import React from "react";
```
iamnotreddy, line 1: Can we rewrite this component so it takes in any child component as a prop? This way it will be reusable for several other use cases.

I'm also thinking we should have some sort of general prop to control the size. Sizing looks pretty solid for the intro page banner, but thinking about other use cases. cc @LuckyHariharan - who wants to use this type of component for the badge details "page"

components/coding/studentPortal/radix/ProgressSlider.tsx
```diff
@@ -0,0 +1,36 @@
+import React, { useEffect, useState } from "react";
+import { Root, Indicator } from "@radix-ui/react-progress";
+
+const ProgressComponent = () => {
+  const [progress, setProgress] = useState(13);
```
iamnotreddy, line 5: thinking initial state could take whatever prop we pass into it

tailwind.config.js
```diff
@@ -98,14 +98,47 @@ module.exports = {
           },
           "100%": { opacity: "1", transform: "scale(1)" },
         },
+        slideDownAndFade: {
```
iamnotreddy, line 101: 🥳

iamnotreddy (changes requested):
Spoke with vithushan earlier today and we're cool moving forward with your approach to implement radix in the codebase. Awesome work on this! Impressed with how fast you did this exploration and I'm really liking how it looks so far. 

Let's actually push this PR through after a couple changes:
- let's remove the implementation from main skillify pages / components for now (sidebar, index, studentPortal etc.)
- let's just push the dialog, progressSlider, and tooltip as general, reusable components into the `components/ui/` directory
- let's make all of these components _reusable_  - they should take in different props: child components, sizing options, values (for slider) etc. so we can use them anywhere across Skillify. no need to be perfect or fit all use cases on the first try... cc @LuckyHariharan who has a good first use case
- let's make all of these storybook components... once you get them polished up, we can create a storybook component for each one. I actually haven't done this yet so we can learn how to do it together

PR #888, opened 2023-02-24
Reroute students from coachingdashboard

LuckyHariharan:
[In this small PR I completed this asana ticket to reroute students from coachingdashboard to the studentPortal dashboard](https://app.asana.com/0/1201390044939970/1203848752604168)

### DIFF
--- a/pages/coaching-dashboard.tsx
+++ b/pages/coaching-dashboard.tsx
@@ -52,7 +52,7 @@ const coachingDashboard = () => {
       return <div>Loading...</div>;
     },
     onCompleted: (roleData) => {
-      if (roleData.users[0].userRole.value === "student") {
+      if (roleData.users[0].userRole.value !== "coach") {
         router.replace("/studentPortal");
       }
     },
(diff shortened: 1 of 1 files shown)


pages/coaching-dashboard.tsx
```diff
@@ -38,6 +44,19 @@ const coachingDashboard = () => {
...
   );
+  useQuery<FetchUserRoleData>(FETCH_USER_ROLE, {
+    variables: {
+      _id: user.uid,
+    },
+    if(loading) {
+      return <div>Loading...</div>;
+    },
+    onCompleted: (roleData) => {
+      if (roleData.users[0].userRole.value === "student") {
```
vithushan19, line 55: Can we change this to check if the value is not equal to "coach" since @kwiedenhaupt is adding a new freemiun role and we wouldn't want them to have access to the coaching dashboard either.

vithushan19 (approved):
This is child's play for you! Left a small comment to address.

pages/coaching-dashboard.tsx
LuckyHariharan: oh good point! will do.

PR #890, opened 2023-02-24
Add a theme switcher to the top of the student portal

vithushan19:
This PR is a large refactor. Instead of using tailwind dark: prefixes, we're going to support a default and a dark theme. This will allow the user to toggle between the two themes. It will also allow for new themes to be supported in the future.

Several components required refactoring to use the new tailwind themed colors.

### DIFF
--- a/styles/globals.css
+++ b/styles/globals.css
@@ -15,4 +15,18 @@
   src: url("/fonts/Poppins/Poppins-Bold.ttf");
 }
 
+.theme-default {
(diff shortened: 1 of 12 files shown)


styles/globals.css
```diff
@@ -15,4 +15,20 @@
   src: url("/fonts/Poppins/Poppins-Bold.ttf");
 }
 
+.theme-default {
```
vithushan19, line 18: theme definitions

PR #896, opened 2023-02-26
ProgressFreemiumComponent and storybook

kwiedenhaupt:
In this PR I did the following: 
1. Added ProgressFreemiumComponent
2. created assocaited storybook
3. added a line to the storybook script file 
https://github.com/webpack/webpack/issues/14532

THIS WAS THE ERROR:
web/node_modules/loader-runner/lib/LoaderRunner.js:205:4 {
  opensslErrorStack: [ 'error:03000086:digital envelope routines::initialization error' ],
  library: 'digital envelope routines',
  reason: 'unsupported',
  code: 'ERR_OSSL_EVP_UNSUPPORTED'
}

image

This component will be used in the Sidebar to display #daysCompletedInTrial/TotalTrialDays under the user's avatar.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,1 @@
-{
-"editor.codeActionsOnSave": {
-    "source.fixAll": true,
-    "source.organizeImports": true
-    }
-}
+{}
--- a/components/ui/ProgressFreemiumComponent.tsx
+++ b/components/ui/ProgressFreemiumComponent.tsx
@@ -1,0 +1,39 @@
+import { Indicator, Root } from "@radix-ui/react-progress";
+import React, { useEffect, useState } from "react";
+
+export interface ProgressFreemiumComponentProps {
+  elapsedDays: number;
+  totalTrialDays: number;
+}
+const ProgressBarComponent: React.FC<ProgressFreemiumComponentProps> = ({
+  elapsedDays,
+  totalTrialDays,
+}) => {
+  const [progress, setProgress] = useState(0);
+  useEffect(() => {
+    setProgress(elapsedDays);
+  }, []);
+
+  return (
+    <div>
+      <Root
+        className="relative overflow-hidden bg-white rounded-full w-4/5 h-[8px]"
+        style={{
+          transform: "translateZ(0)",
+        }}
+        value={progress}
+      >
+        <Indicator
+          className="bg-rattata w-full h-full transition-transform duration-[660ms] ease-[cubic-bezier(0.65, 0, 0.35, 1)]"
+          style={{
+            transform: `translateX(-${
+              100 - (progress / totalTrialDays) * 100
+            }%)`,
+          }}
+        />
+      </Root>
+    </div>
+  );
+};
+
+export default ProgressBarComponent;
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "@hotjar/browser": "^1.0.6",
     "@magic-ext/oauth": "^2.1.1",
     "@matejmazur/react-katex": "^3.1.3",
+    "@radix-ui/react-progress": "^1.0.1",
     "@radix-ui/react-tooltip": "^1.0.3",
     "@react-spring/web": "^9.6.1",
     "@react-three/drei": "^7.27.5",
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "lint": "next lint",
     "start": "next start",
     "test": "env TZ='UTC' jest",
-    "storybook": "start-storybook -s ./public -p 6006",
+    "storybook": "export NODE_OPTIONS=--openssl-legacy-provider && start-storybook -s ./public -p 6006",
     "build-storybook": "build-storybook -s public",
     "chromatic": "npx chromatic --project-token=3849dc6eb670"
   },
--- a/components/ui/ProgressFreemiumComponent.stories.tsx
+++ b/components/ui/ProgressFreemiumComponent.stories.tsx
@@ -1,0 +1,20 @@
+import { Meta, Story } from "@storybook/react";
+import ProgressFreemiumComponent, {
+  ProgressFreemiumComponentProps,
+} from "./ProgressFreemiumComponent";
+import React from "react";
+
+export default {
+  title: "UI/Progress",
+  component: ProgressFreemiumComponent,
+} as Meta;
+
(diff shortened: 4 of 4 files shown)


components/ui/ProgressFreemiumComponent.tsx
```diff
@@ -0,0 +1,39 @@
+import { Indicator, Root } from "@radix-ui/react-progress";
+import React, { useEffect, useState } from "react";
+
+export interface ProgressFreemiumComponentProps {
+  elapsedDays: number;
+  totalTrialDays: number;
+}
+const ProgressBarComponent: React.FC<ProgressFreemiumComponentProps> = ({
```
vithushan19, line 8: Component name and prop type name don't match

components/ui/ProgressFreemiumComponent.tsx
```diff
@@ -0,0 +1,39 @@
...
+  elapsedDays,
+  totalTrialDays,
+}) => {
+  const [progress, setProgress] = useState(0);
+  useEffect(() => {
```
vithushan19, line 13: Why do you need the state and effect hook? If you're not changing state you can just directly use the prop that you're passing in

.vscode/settings.json
```diff
@@ -1,6 +1 @@
-{
```
vithushan19, line 1: Are you sure we want to remove this for others? I found it quite helpful in reducing on used imports. I know it gave you an issue but maybe there's something else going on that we can sync on Tues to figure out?

components/ui/ProgressFreemiumComponent.tsx
```diff
@@ -0,0 +1,39 @@
...
+    setProgress(elapsedDays);
+  }, []);
+
+  return (
+    <div>
+      <Root
+        className="relative overflow-hidden bg-white rounded-full w-4/5 h-[8px]"
```
vithushan19, line 20: What's the reasoning for using 4/5 instead of absolute widths? Did you check how this looks on different screen sizes?

vithushan19: Couldn't you use one of the default tailwind heights instead of hard coding 8px? I think one of the classes is equivalent to 8px I think it's h-2

package.json
```diff
@@ -8,7 +8,7 @@
     "lint": "next lint",
     "start": "next start",
     "test": "env TZ='UTC' jest",
-    "storybook": "start-storybook -s ./public -p 6006",
+    "storybook": "export NODE_OPTIONS=--openssl-legacy-provider && start-storybook -s ./public -p 6006",
```
vithushan19, line 11: Are we sure we need this? I'm ok to merge this in but I want to discuss further on a call

vithushan19 (commented):
Some questions and feedback

components/ui/ProgressFreemiumComponent.tsx
kwiedenhaupt: changed to match

components/ui/ProgressFreemiumComponent.tsx
kwiedenhaupt: great point.  I am now relying on the prop elapsedDays

components/ui/ProgressFreemiumComponent.tsx
kwiedenhaupt: The reasoning is that the current use will be in the Sidebar component, which will force the size on both mobile and computer to be limited to the sidebar width.  

Great point concerning the tailwind classes.  I used h-2 and w-40.  Checked that on mobile and computer as well.

package.json
kwiedenhaupt: I was able to get it to run without it.  Seems to be fixed!

.vscode/settings.json
kwiedenhaupt: That makes a lot of sense.  I am curious as to how to keep the react imports while using this script.  The react imports are required to take care of the lint errors, correct? 

Am interested to discuss this at the standup tomorrow.

components/ui/ProgressFreemiumComponent.tsx
```diff
@@ -0,0 +1,34 @@
+import { Indicator, Root } from "@radix-ui/react-progress";
```
iamnotreddy, line 1: I could see us using this type of component across the Skillify site - thinking badge completion progress or even for course completion

What if we generalize this so it's not just for Freemium? Perhaps we could call it ProgressComponent and it could take in two props like currentValue and totalValue or something general like that

components/ui/ProgressFreemiumComponent.tsx
```diff
@@ -0,0 +1,34 @@
...
+        style={{
+          transform: "translateZ(0)",
+        }}
+        value={elapsedDays}
+      >
+        <Indicator
+          className="bg-rattata w-full h-full transition-transform duration-[660ms] ease-[cubic-bezier(0.65, 0, 0.35, 1)]"
+          style={{
+            transform: `translateX(-${
+              100 - (elapsedDays / totalTrialDays) * 100
```
iamnotreddy, line 25: don't think I fully understand how this works: 

you're finding the whole number percentage value and subtracting it from 100 to get the "progress remaining"?

iamnotreddy (changes requested):
I think there's an opportunity to generalize this component further (see below) but lmk if you feel otherwise. 

I tried calling the component on a random page and I noticed it's hard to visually distinguish how much progress has been completed and how much is remaining:
Screen Shot 2023-02-27 at 2 15 27 PM

I'm wondering if it makes sense to put a border around the bar and / or shade the "remaining" section a very very subtle gray-ish color to make it more clear visually... lmk what you think

Nice work!

components/ui/ProgressFreemiumComponent.tsx
kwiedenhaupt: I just changed it.  great idea.  the initial reason I didn't do this was because we had a ProgressBar.tsx, which is really the same thing.  I found it after I created this one, when we moved it into the UI folder.  I didn't want it to become confusing.  So I think the names you proposed are much better -- I just added a comment to distinguish between them.  let me know if you think that's a good idea.

components/ui/ProgressFreemiumComponent.tsx
kwiedenhaupt: That's exactly how it works.

iamnotreddy (approved):
nice quick work on the changes!

PR #897, opened 2023-02-27
Dialog Component and Story

kwiedenhaupt:
In this PR I created a reusable dialog component, along with a story.
This component takes in children (the content) as well as a number of props.  
I used the standard Skillify button, however, the bg-brandPrimary style conflicted with my component, and so I had to modify the Button component to accept a manual bg-charmander background that does not conflict. 

In order to deal with the conflicting styles I could: 
1. create a new button with similar classes
2. copy the classes for the button and put them in a div surrounding the button to override the classes in the button
3. choose a separate color for the button
4. modify the button with bg-charmander manually typed out instead of using bg-brandPrimary.

I am open to completely changing the styling of the button, if anyone thinks that would be cleaner or cooler, and then would remove the changes on our Skillify Button.  I'd like to have feedback regarding this styling as well as other styling for this box from other team members, as I might not be thinking of all the use cases that are most likely.

Here is the story for the component:
image


components/ui/Button.tsx
```diff
@@ -13,7 +13,8 @@ export interface ButtonProps {
     | "pink"
     | "yellow"
     | "orange"
-    | "white";
+    | "white"
+    | "DialogComponent";
```
vithushan19, line 17: We should be able to use the primary brand colour here, if that's not working we can discuss on a call

components/ui/DialogComponent.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const Template: Story<DialogComponentProps> = (args) => (
+  <DialogComponent {...args}>
+    <div className="w-40 cursor-pointer">
+      <p className="text-center">This is the Content.</p>
+    </div>
+  </DialogComponent>
+);
+
+export const Primary = Template.bind({});
```
vithushan19, line 18: You're looking like a storybook pro, when you merge your PRs in ask me to unlock a storybook badge for you.

components/ui/DialogComponent.tsx
```diff
@@ -0,0 +1,79 @@
...
+  trigger: boolean;
+  message?: string;
+  link: string;
+  buttonLabel: string;
+  header?: string;
+  triggerTitle?: string;
+  size: string;
+}
+
+const DialogComponent: React.FC<DialogComponentProps> = ({
```
vithushan19, line 24: Beautiful ❤️

vithushan19 (commented):
Looks good to me, think we should sync on the new styling convention for the button before merging

components/ui/Button.tsx
kwiedenhaupt: Yeah.  I'd love to sync on that, because I want to see the button component work inside this component, and take care of the conflicting styles.  I'm def interested in troubleshooting that!

redux/themeSlice.ts
```diff
@@ -0,0 +1,35 @@
...
+  DEFAULT = "theme-default",
+  DRACULA = "theme-dracula",
+}
+
+export type ThemeState = {
+  currentTheme: Theme;
+};
+
+const initialState: ThemeState = {
+  currentTheme: Theme.DRACULA,
```
iamnotreddy, line 14: @vithushan19 - I think we should set the initial theme to light mode (DEFAULT) instead of dark (DRACULA). 

I noticed that there's a few components in dark mode that don't look good and need restyling:
- coaches page cards
- expandable containers 
- goals (text blends in with dark bg)

pages/studentPortal/index.tsx
```diff
@@ -67,6 +68,15 @@ export default function StudentPortalPage() {
 
   return (
     <div className="flex flex-col w-full px-4 pb-4 sm:px-8 sm:pb-8 ">
+      <DialogComponent
```
iamnotreddy, line 71: let's take this component out for now until we have logic to call it conditionally (i.e you're first time on skillify, trial is over etc.)

iamnotreddy (approved):
awesome job on this

two small things to address before merging but otherwise looks great

PR #898, opened 2023-02-27
Fix routes for admin and profile

LuckyHariharan:
Hey Vithushan, put up a PR to fix the routing for the coaching dashboard and profiles. Also non-coaches are rerouted from the coaching dashboard as it did previously.

- /studentPortal navigates to the dashboard
- /profile/[userid] navigates to profile page
- /profile/admin navigates to the coaching-dashboard
- clicking on a card in the coaching dashboard navigates to profile/[userid]


components/studentPortal/layout/Sidebar.tsx
```diff
@@ -192,7 +192,7 @@ export const Sidebar: React.FC = () => {
         {userRole === "coach" && (
           <SidebarItem
             name={"Admin"}
-            link={"/studentPortal/admin"}
+            link={"/profile/admin"}
```
vithushan19, line 195: From a code organization perspective, I think the admin page should be part of the student portal not the profile

vithushan19 (commented):
Instead of changing the routes, can we move the pages so that the admin page is part of the portal?

PR #899, opened 2023-02-27
Fix link for profile cards

vithushan19:


pages/studentPortal/admin/index.tsx
```diff
@@ -121,7 +120,7 @@ const coachingDashboard = () => {
           const completionDate = goalCompletionDateList[index];
           const completedGoal = completedGoalsList[index];
           return (
-            <Link href={"profile/" + it.id} key={index}>
+            <Link href={"/profile/" + it.id} key={index}>
```
LuckyHariharan, line 123: learned something about next.js today!

PR #900, opened 2023-02-28
Expandable Container Storybook

kwiedenhaupt:
image

In creating the storybook I am wondering to make it more reusable: 
1. Would a width prop be a good idea?
2. how about a color prop for the container/title?

I also have a general question on stories: is the information populating the storybook important?  i.e. should I pull a query from the database in to show that it's height can be relative and will change, etc?
Are there ways I should be thinking about to make this better?


vithushan19 (approved):
this looks good to me. I need to do some investigation into these expandable containers. I actually took them out of the profile for now. 

When we had multiple containers on the profile page, we were running into issues where scrolling up and down on the profile would show a glitchy blank screen for a second, while React took it's time re-rendering. 

Once I look into why that rendering issue was happening, we can re-integrate this container back into the profile. As far putting in this into our storybook, im happy with this pr and it can be merged in.

PR #904, opened 2023-03-03
Build User Selection UI - Skillify Career Quiz

LuckyHariharan:
In this PR, I built the UI for the page. This is my first draft, and although I think it mirrors the spec I would love feedback. 

This component lives in pages/studentPortal/quizcomponent for easy viewing. The plan is to store it somewhere in the component folder later. @vithushan19 please lmk your thoughts. 

image
image


pages/studentPortal/quizcomponent/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+type SkillSelectionProps = {
+  selections: string[];
+};
+
+const SkillSelection: React.FC<SkillSelectionProps> = ({ selections }) => {
+  const [selected, setSelected] = useState<string[]>([]);
+
+  const handleSelection = (selection: string) => {
+    if (selected.includes(selection)) {
+      setSelected(selected.filter((s) => s !== selection));
```
vithushan19, line 12: nice use of filter and spread operator to support adding and removing from the list. I think it's good for now, but we may want to move this into redux later once we get more components. Perhaps redux will be overkill, so we can keep this for now and see how the feature progresses. Just something you should think about

pages/studentPortal/quizcomponent/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+    <div className="flex flex-col w-full max-w-4xl mx-auto p-4">
+      <div className="w-full mb-4 text-center">
+        <div className="font-bold text-2xl text-black-600"></div>
+      </div>
+      {selections.map((selection, index) => (
+        <div
+          key={index}
+          className={`flex   items-start justify-start w-full px-4 py-2 my-2  cursor-pointer ${
+            selected.includes(selection)
+              ? "bg-violet-300 text-black-500 border-2 border-black-500 rounded-[14px]"
```
vithushan19, line 28: the border-2 border-black-500 rounded-[14px] could be move outside of the ternary. This would make it easier to understand how the the ternary condition affects the style.

Also, I'm not big on using customized hard-coded values in tailwind styles, unless necessary. Do the pre-defined rounded tailwind classes not work for you? like rounded-lg?

pages/studentPortal/quizcomponent/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+      ))}
+    </div>
+  );
+};
+
+type ProgressBarProps = {
+  progress: number;
+};
+
+const ProgressBar: React.FC<ProgressBarProps> = ({ progress }) => {
```
vithushan19, line 44: could use of extracting components out into different functions, as the feature grows, it would be good to have each component live in it's own file all inside the same folder.

vithushan19 (approved):

pages/studentPortal/quizcomponent/index.tsx
LuckyHariharan: Yeah they do work, I'll use the defaults.

pages/studentPortal/quizcomponent/index.tsx
LuckyHariharan: Understood. Will organize my components before merging.

PR #905, opened 2023-03-06
Setting up Sidebar for Freemium Role

kwiedenhaupt:
image
medium screens
mobile view
image

Main Changes:

Database:
1.  freemium role was added to the roles enum.
2. default role is set to "freemium" and so when a student enrolls, they will have to be manually changed.

GraphQL queries:
1.  Modified fetchUserRole query to include created_at value so that no additional network calls needed to be made.  
2. Checked it's use in other places in the database.  They are not affected, but changes to the profileSlice and the associated types did need to be made and were made.

Small Bug Fix
1. the "admin", "web dev", "coding basics" pages were not included in setActive page list -- added that

General UI Change
image
1. implemented svg icons to increase the professional look.

Freemium User Sidebar:
1.  removed the avatar from the mobile view because otherwise the name and title were encroaching onto the main view when the sidebar was collapsed.
2. implemented ProgressComponent
    a.  created a functions page in the API folder named freemium.ts that hold the elapsedDays and trialDaysRemaining 
    functions that will be used in the various components.
3.  conditionally rendered the view of the sidebar to be grayed out and sized appropriately for the freemium user.
4. disabled all links but coding basics and web development for freemium users.

NOTES:
1.  I tried my best to adhere to DRY principles, although it seems that for the freemium user view code does need to be duplicated.  Suggestions are welcome concerning the way I approached that with the sidebar, as I don't want bloat especially in this component. 
2. I chose to store created_at in redux in the profileSlice due to the fact that I will be using it for the other components such as DialogComponent that will be rendered on different pages of the app.  open to feedback.
3. Should the hover functionality work over grayed out features?  I left it but wasn't sure what was best. 
4. Open to feedback on the icon/sizing for the freemium user role.  I thought a bigger avatar for md and no avatar for mobile was best.  I chose to specialize it because with the additional progress bar, etc. it had to be changed from the original anyway and therefore was not just repeating code merely for cosmetics.


components/studentPortal/layout/Sidebar.tsx
```diff
@@ -22,45 +31,52 @@ interface SidebarItemProps {
   page: SidebarPage;
   icon: ReactElement;
   notifications?: boolean;
+  userRole: "freemium" | "student" | "coach";
 }
 const SidebarItem = ({
```
vithushan19, line 36: I think it's time to move sidebarItem into it's own file now

components/studentPortal/layout/Sidebar.tsx
```diff
@@ -22,45 +31,52 @@ interface SidebarItemProps {
...
   name,
   link,
   page,
   icon,
   notifications,
+  userRole,
 }: SidebarItemProps) => {
   const { activePage } = useSelector(activePageSelector);
+  const isDisabled = userRole === "freemium";
```
vithushan19, line 45: we should just pass isDisabled in as a prop. There's no need for the sidebarItem to know about userRoles. The component just needs to know if it should render itself as disabled or not

components/studentPortal/layout/Sidebar.tsx
```diff
@@ -22,45 +31,52 @@ interface SidebarItemProps {
...
 };
 
 export const Sidebar: React.FC = () => {
   const { goalApproaching } = useSelector(activePageSelector);
-  const { userRole } = useSelector(profileSelector);
+  const { userRole, createdAt } = useSelector(profileSelector);
   const dispatch = useDispatch();
-
   const { signOut, user } = useAuth();
+  const totalTrialDays = 30;
```
vithushan19, line 79: for hardcoded constants like this, the convention is to use TOTAL_TRIAL_DAYS instead

components/studentPortal/layout/Sidebar.tsx
```diff
@@ -69,8 +85,11 @@ export const Sidebar: React.FC = () => {
     onCompleted: (data) => {
       if (data.users[0].userRole.value === "coach") {
         dispatch(setUserRole("coach"));
-      } else {
+      } else if (data.users[0].userRole.value === "student") {
         dispatch(setUserRole("student"));
+      } else if (data.users[0].userRole.value === "freemium") {
+        dispatch(setCreatedAt(data.users[0].created_at));
```
vithushan19, line 91: won't this get set everytime a freemium user logs in?

vithushan19: we can actually create a created_at column in the users table in hasura. and default the value to be `now()` this way, we will set that column, anytime a new user is created and added to the table.

components/studentPortal/layout/Sidebar.tsx
```diff
@@ -96,22 +121,46 @@ export const Sidebar: React.FC = () => {
     //Full width then restrict in page
     <div className="flex flex-col w-full bg-backgroundPrimary text-textPrimary">
       <div className="grid">
-        <div className="flex p-4">
-          {user && (
+        {user && userRole === "freemium" ? (
+          <div className="flex items-center p-4">
```
vithushan19, line 125: I might extract this out into a component called `FreemiumSidebarHeader`

components/studentPortal/layout/Sidebar.tsx
```diff
@@ -96,22 +121,46 @@ export const Sidebar: React.FC = () => {
...
+              <p className="text-xs mt-1 text-gray-500">
+                {trialDaysRemaining(createdAt, totalTrialDays)}/{totalTrialDays}{" "}
+                days remaining
+              </p>
             </div>
-          )}
-        </div>
+          </div>
+        ) : (
+          <div className="flex p-4">
```
vithushan19, line 147: this could be extracted to `SidebarHeader`

components/studentPortal/layout/Sidebar.tsx
```diff
@@ -126,6 +175,7 @@ export const Sidebar: React.FC = () => {
               <path d="M5 3a2 2 0 00-2 2v2a2 2 0 002 2h2a2 2 0 002-2V5a2 2 0 00-2-2H5zM5 11a2 2 0 00-2 2v2a2 2 0 002 2h2a2 2 0 002-2v-2a2 2 0 00-2-2H5zM11 5a2 2 0 012-2h2a2 2 0 012 2v2a2 2 0 01-2 2h-2a2 2 0 01-2-2V5zM11 13a2 2 0 012-2h2a2 2 0 012 2v2a2 2 0 01-2 2h-2a2 2 0 01-2-2v-2z" />
             </svg>
           }
+          userRole={userRole}
```
vithushan19, line 178: sidebarItems don't need the userRole. UserRole is kind of a sensitive property that we don't want to make accessible everywhere unless absolutely needed. 

Rather than passing in userRole, I would just pass in isDisabled={userRole === "freemium"}

components/studentPortal/layout/Sidebar.tsx
```diff
@@ -222,17 +276,12 @@ export const Sidebar: React.FC = () => {
...
...          <path d="M 43.335938 4 L 6.667969 4 C 5.195313 4 4 5.195313 4 6.667969 L 4 43.332031 C 4 44.804688 5.195313 46 6.667969 46 L 43.332031 46 C 44.804688 46 46 44.804688 46 43.335938 L 46 6.667969 C 46 5.195313 44.804688 4 43.335938 4 Z M 27 36.183594 C 27 40.179688 24.65625 42 21.234375 42 C 18.140625 42 15.910156 39.925781 15 38 L 18.144531 36.097656 C 18.75 37.171875 19.671875 38 21 38 C 22.269531 38 23 37.503906 23 35.574219 L 23 23 L 27 23 Z M 35.675781 42 C 32.132813 42 30.121094 40.214844 29 38 L 32 36 C 32.816406 37.335938 33.707031 38.613281 35.589844 38.613281 C 37.171875 38.613281 38 37.824219 38 36.730469 C 38 35.425781 37.140625 34.960938 35.402344 34.199219 L 34.449219 33.789063 C 31.695313 32.617188 29.863281 31.148438 29.863281 28.039063 C 29.863281 25.179688 32.046875 23 35.453125 23 C 37.878906 23 39.621094 23.84375 40.878906 26.054688 L 37.910156 27.964844 C 37.253906 26.789063 36.550781 26.328125 35.453125 26.328125 C 34.335938 26.328125 33.628906 27.039063 33.628906 27.964844 C 33.628906 29.109375 34.335938 29.570313 35.972656 30.28125 L 36.925781 30.691406 C 40.171875 32.078125 42 33.496094 42 36.683594 C 42 40.117188 39.300781 42 35.675781 42 Z"></path>
```
vithushan19, line 284: these icons are cluttering the sidebar component. maybe extract them into a `JSIcon` and `ReactIcon` component?

graphql/studentPortal/users/fetchUserRole.ts
```diff
@@ -1,18 +1,21 @@
...
 
 export const FETCH_USER_ROLE = gql`
-  query FetchUserRole($_id: String = "") {
-    users(where: { id: { _eq: $_id } }) {
-      userRole {
-        value
-      }
+query FetchUserRole($_id: String = "") {
+  users(where: {id: {_eq: $_id}}) {
+    created_at
```
vithushan19, line 6: this query is fine, but instead of setting this value in the React app, we should just use the default values in hasura to automatically set the created_at property when a new user is created

graphql/studentPortal/users/fetchUserRole.ts
```diff
@@ -1,18 +1,21 @@
...
+
 `;
 
 export type FetchUserRoleData = {
   users: UserRoleData;
 };
 export type UserRoleData = {
-  userRole: "student" | "coach";
+  userRole: "student" | "coach"| "freemium";
+  created_at: Date;
```
vithushan19, line 20: created_at isn't  really part of UserRoleData, I might rename this class to `UserData`

pages/api/studentPortal/freemium.ts
```diff
@@ -0,0 +1,25 @@
+import { addDays, differenceInCalendarDays } from 'date-fns';
+
+//used for freemium feature calculations.
+//createdAt comes from fetchUserRole query and trialLength is manually inputted.
+export const trialDaysRemaining = (createdAt:Date, trialLength:number) => {
```
vithushan19, line 5: rather than putting another file in the studentPortal folder, lets create a subfolder called `freemium` and rename this to index.ts. Or you could break these functions into two different files and name the files after the function names. I prefer this latter approach.

Then you can also create unit tests for both functions. 

These functions are good candidates to create unit tests for. and creating a folder for freemium will isolate all the work you're doing from all the other student portal work.

vithushan19: it's always a good idea to test your api functions

tsconfig.json
```diff
@@ -12,7 +16,7 @@
     "moduleResolution": "node",
     "resolveJsonModule": true,
     "isolatedModules": true,
-    "jsx": "react",
+    "jsx": "preserve",
```
vithushan19, line 19: do you want to make this change, isn't this what messed with your eslint autofixes?

redux/profileSlice.ts
```diff
@@ -70,6 +72,18 @@ export const profileSlice: Slice = createSlice({
         };
       }
     },
+
+    setCreatedAt: (
```
vithushan19, line 76: i think this should be good to leave in. we just want hasura to automatically set the created_at value, and then you can query it and set it here like you're already doing

vithushan19 (commented):
great work! you put together a really great PR. super detailed and clear, amazing to see your growth. this is such a heavy feature, and you really broke it down well.

Now I'm able to give you a deeper level of feedback around code readability and maintainability. You can already piece things together to make code work, but can you write code in an elegant way that's easy for others to read.

components/studentPortal/layout/Sidebar.tsx
kwiedenhaupt: did that. good suggestion.  the delete unused imports functionality you implemented works great for this sort of thing.

components/studentPortal/layout/Sidebar.tsx
kwiedenhaupt: done. used a state hook to set it in the Sidebar.tsx

components/studentPortal/layout/Sidebar.tsx
kwiedenhaupt: no. this is the date that the account was created at.  And it is a column in the users table, and I am adding it into this query and it should be set to now ()

graphql/studentPortal/users/fetchUserRole.ts
kwiedenhaupt: That is what's done.  I'm just reading it here from the hasura database.

tsconfig.json
kwiedenhaupt: I keep rejecting these changes and it keeps automatically making them... maybe you can help me sometime figure this out.

components/studentPortal/layout/Sidebar.tsx
kwiedenhaupt: great idea.  Created a freemium folder in the studentPortal section of components and put it there.
makes the code a lot cleaner in the sidebar.

pages/api/studentPortal/freemium.ts
kwiedenhaupt: I created two unit test suites:
trialDaysRemaining.test.ts and elapsedDays.test.ts

Two questions: 
What do I want to return in each function if the created_at date is accidentally modified in the database and it ends up being a date after the current system date?  

I chose to have the trialDaysRemaining function return 0 and the elapsedDays function return 0 --- but wasn't sure if there was a reason to have an undefined so it shows up as NaN??

components/studentPortal/layout/Sidebar.tsx
vithushan19: my mistake, i didn't understand what was happening. looks good tome

vithushan19 (approved):
nice work @kwiedenhaupt you should be really proud that you're able to write code like this. You were able to create this level of code quality, with just a single review from me. Even after years of experience, there is always an opportunity to improve your code after a code review. 

It really feels like you're thinking in React and understanding how things connect. I would think about opportunities for you to move across the stack more. Seems like you can create an UI experience that could be designed, and now you need more experience working with data and backend. 

You should sync with Grace and make sure that you have a good understanding of what the rest of the freemium experience will look like, but once we ship this feature lets think of ways for you to grow in other parts of the tech stack!

PR #910, opened 2023-03-11
Grayed out content for Web Dev Freemium users

kwiedenhaupt:
Freemium View:
image

Student View:
image

Overall Goal:
1. To gate content for the freemium role only: gray out/deactivate links for content after the "Components and Props" lesson in react.
2. To add a node that is a plug for Skillify

Changes made:
1. used redux to call the userRole variable into the Web Development index page.
2. drilled down prop for userRole to UnitNodeView and UnitView and made all changes dependent on the userRole === "freemium"
3. created an arrays freemiumUnits and freemiumNodes that contain units/nodes that are included in freemium access.
4. created a prop "grayedOut" to control the nodes that were to be displayed
image
5. created a prop "freemiumMessage" in order to display the plug for skillify at the node directly after the active content.
6. Linked the Apply Now! button to booking a call with Vithushan.

Visual changes:
image
Grayed out rounded divs are displayed in the place of the unit titles and node titles/descriptions/links.

No current database interaction on these changes.

PLEASE NOTE:
An alternate approach should be considered, which would include adding all the nodes to the db and adding a freemium boolean column and adding that information to the querys, the Units.tsx, the other components and the transform function.


pages/studentPortal/web/index.tsx
```diff
@@ -42,7 +44,7 @@ export default function StudentPortalPage() {
       />
       <div className="grid grid-cols-1 gap-4">
         {units.map((it, index) => (
-          <UnitView key={index} data={it} />
+          <UnitView key={index} data={it} userRole={userRole} />
```
vithushan19, line 47: You did a great job implementing this feature but we should discuss higher level code design.

An important principle for you to understand is the single responsibility principle.

You want each component to be responsible for just one thing. You want to try and separate your UI code from the logic code.

The unit view is a UI component that should only have to worry about whether to render itself as locked, enabled, grayed out etc...

The UI should not need to know anything about the currently logged in user.

Your implementation requires the unitview depend on the current users role.

Let's discuss how we can change your code design so your UI component doesn't depend on user role.

pages/studentPortal/web/index.tsx
```diff
@@ -1,16 +1,18 @@
...
 import { FETCH_USER_INTRO_NODES } from "../../../graphql/studentPortal/courses/fetchUserIntroNodes";
 import { UPDATE_USER } from "../../../graphql/studentPortal/users/updateUser";
 import { useAuth } from "../../../lib/authContext";
+import { profileSelector } from "../../../redux/profileSlice";
 import { reactUnits, Unit } from "../../api/studentPortal/units";
 
 export default function StudentPortalPage() {
   const { user } = useAuth();
-
+  const { userRole } = useSelector(profileSelector);
```
vithushan19, line 15: It's fine for the page to need to know about the user role, but our lower level components should not.

vithushan19: The page is the higher level orchestrator that connects everything.

components/studentPortal/lessons/UnitView.tsx
```diff
@@ -4,39 +4,65 @@ import UnitNodeView from "./UnitNodeView";
 
 export type UnitViewProps = {
   data: Unit;
+  userRole: "freemium" | "student" | "coach";
```
vithushan19, line 7: We don't want the unit view to depend on user role

components/studentPortal/lessons/UnitView.tsx
```diff
@@ -4,39 +4,65 @@ import UnitNodeView from "./UnitNodeView";
...
 };
 
-export const UnitView: React.FC<UnitViewProps> = ({ data }: UnitViewProps) => {
+//These are hardcoded lists of the Units/Nodes that will be included with the Freemium version
+const freemiumUnits = ["Introduction", "Github", "TailwindCSS", "React"];
+const freemiumNodes = [
```
vithushan19, line 12: The unit view component should only have to know how to render itself. It shouldn't have to know what units make up the freemium curriculum.

The curriculum info is part of our business' logic. We need to separate that from UI component code.

pages/studentPortal/web/index.tsx
```diff
@@ -42,7 +44,7 @@ export default function StudentPortalPage() {
       />
       <div className="grid grid-cols-1 gap-4">
         {units.map((it, index) => (
```
vithushan19, line 46: Maybe you can check the user role above this line. Show the default units for most students, but if role equals freemium we show a FreemiumCourseComponent.

This may not work but something we should think about together.

vithushan19 (commented):
Nice work. This code is functionally correct, but we need to refector it to make it easier to maintain.

If we wanted to scrap the freemium feature tomorrow, it would be very difficult to undo your work without breaking things or causing bugs.

Ideally we would have a new freemiun folder of components that could easily be deleted.

The unit components just need to be changed to support graying out.

Everything else should ideally be in a freemium folder.

pages/studentPortal/web/index.tsx
kwiedenhaupt: This makes a lot of sense.  
I have removed the userRole from all the components, and check it in the index.tsx as you suggest and then deploy a freemium or a regular experience depending on the role as called on that page.

I moved the curriculum/business logic to a freemiumUnits.ts file so that this controls the units that we want to include or not include and deploys them the same way to maintain consistency between how we are doing things with the freemium users and the full users.

image

This is one solution, but I would love to chat to see other ideas that you would suggest to either scrap this approach or improve on it!

pages/studentPortal/web/index.tsx
kwiedenhaupt: removed userRole from other components.  Created freemium components instead and checked userRole on the index page and deployed components accordingly from there.

pages/studentPortal/web/index.tsx
kwiedenhaupt: I have done this but broken it into two components like the original.  Is this too much code duplication or do you think it works?

components/studentPortal/lessons/UnitView.tsx
kwiedenhaupt: I created a freemiumUnits.ts file that includes only the units that we want to include with freemium and includes a few empty units to gray out.  This can be easily changed and does not affect the regular user.

components/studentPortal/freemium/FreemiumUnitNodeView.tsx
```diff
@@ -0,0 +1,129 @@
...
+  completed,
+  locked,
+  hiddenLine,
+  title,
+  description,
+  type,
+  grayedOut,
+  freemiumMessage,
+}: FreemiumUnitNodeViewProps) => {
+  const imageSrc = (completed, locked, type, freemiumMessage) => {
```
vithushan19, line 25: Could we extract this UI logic into a shared component that can be used by both the freemium and default unit view component.

Maybe called NodeIcon

components/studentPortal/freemium/FreemiumUnitNodeView.tsx
```diff
@@ -0,0 +1,129 @@
...
+        </div>
+        {grayedOut === true ? (
+          <div className="flex flex-col justify-center space-y-2 w-full col-span-10 ml-4 sm:ml-0 sm:col-span-6">
+            <div className="bg-gray-300 h-6 w-2/3 rounded-full"></div>
+            <div className="bg-gray-300 h-6 w-1/2 rounded-full"></div>
+          </div>
+        ) : (
+          <div className="flex flex-col justify-center w-full col-span-10 ml-4 sm:ml-0 sm:col-span-6">
+            {freemiumMessage ? (
+              <div className="">
```
vithushan19, line 79: Could this be extracted into a component called freemium row?

components/studentPortal/freemium/FreemiumUnits.tsx
```diff
@@ -0,0 +1,136 @@
+import { Unit } from "../../../pages/api/studentPortal/units";
+
+export const freemiumUnits: Unit[] = [
```
vithushan19, line 3: This is great! Much cleaner

vithushan19 (commented):
Left a few more comments, happy to discuss in person

components/studentPortal/freemium/FreemiumUnitNodeView.tsx
kwiedenhaupt: good idea.  put the logic in the api folder and then created a NodeIcon component used by both freemium and regular students.

vithushan19 (approved):
Great work!

PR #911, opened 2023-03-14
Freemium Header

kwiedenhaupt:
Regular View:
image
Mobile: 
image
Figma Sketch:
image

Purpose: 
To create a customized header for the Freemium Component that includes a call to action to sign up that appears on the layout (every page).

1. Created a FreemiumHeader.tsx component that 
2. Check for the userRole/createdAt on the layout page and conditionally render the header components based on the userRole.

UI notes:
1. lined up the buttons for the apply now and the continue for the lessons...this was not the case in the sketches, but it seemed to look the best to me.  Thoughts? 
2. For the mobile view I removed the progress bar, etc. and left only the call to action -- button saying apply now!  I like this but am open to suggestions!
3. Kept the toggle button on the far right to toggle the themes.

Advice Needed: 
1. How do I get the "Enjoying the Skillify Experience" to show up white or grey when the theme is switched to dark?  I've noticed in dark mode a lot of the sidebar, etc. is not correct.  I did use dark: bg-white, but that doesn't work, as I'm sure I'm not fully understanding the interaction between tailwind and the dracula theme.

### DIFF
--- a/components/studentPortal/freemium/FreemiumHeader.tsx
+++ b/components/studentPortal/freemium/FreemiumHeader.tsx
@@ -1,0 +1,66 @@
+import React from "react";
+import { elapsedDays } from "../../../pages/api/studentPortal/freemium/elapsedDays";
+import { trialDaysRemaining } from "../../../pages/api/studentPortal/freemium/trialDaysRemaining";
+import { Theme } from "../../../redux/themeSlice";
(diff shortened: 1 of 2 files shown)


vithushan19 (approved):
nice, you figured out how to use the themes with bg-backgroundPrimary!

PR #917, opened 2023-03-17
freemium prop fixes

kwiedenhaupt:
Purpose of the PR: Clean up the workflow for the freemium feature.

1.  Cleaned up nodeIcon.ts 
- renamed file to imageSrc.ts
- removed the freemiumImageSrc function and combined it with the imageSrc

2. Cleaned up the NodeIcon.tsx
- removed freemiumMessage and description props
- drilled down type prop to the NodeIcon and used type === "freemiumMessage" and type === "grayedOut" to control the rendering of the icon. 

3. Investigated hiddenLine prop 
- this is necessary in order to skip the gray line if the node is the bottom node.  It does look better this way.
- used this hiddenLine feature as a prop in SkeletonNodeView and FreemiumMessageNodeView and made it data.nodes.length-1 to get the last node and make that a "hiddenLine" so the gray line does not show up at the end.

4. Merged FreemiumUnitView.tsx into UnitView.tsx.
This allows for less code duplication as well as keeps the separation for the freemium feature.  It would only be removing two if statements to rip out the feature on this file.


components/studentPortal/lessons/UnitView.tsx
```diff
@@ -14,11 +16,21 @@ export const UnitView: React.FC<UnitViewProps> = ({ data }: UnitViewProps) => {
...
+            return (
+              <FreemiumMessageNodeView
+                hiddenLine={index === data.nodes.length - 1}
+                type={it.type}
+              />
+            );
+          } else if (it.type === "grayedOut") {
+            return (
+              <SkeletonNodeView
+                hiddenLine={index === data.nodes.length - 1}
```
vithushan19, line 29: What does hidden line do?

vithushan19 (approved):
Wow so much cleaner! Nice job!

PR #918, opened 2023-03-20
Modals

kwiedenhaupt:
Goal of PR:
Create 5 modals that appear upon login only once in a user "session" and only for paid and freemium users.

Modal 1
image
Mobile:
image

Modal 2
image
Mobile: 
image

Modal 3
image
Mobile:
image

Modal 4
image
Mobile: 
image

Modal 5
image
Mobile: 
image

NOTES:
1. Created a FreemiumDialogComponent that uses the Dialog component from the Radix library and is open upon load.  This component maps to the five stages from the ModalStage enum
2. Added an overlay that transitions from gray to transparent.  Can change to transparent if preferred.
3. Added TrailAnimation.tsx file from the Spring library that handles transitions between the modals. Saved in the UI folder.
4. Created a function called showModal that checks if the userRole is paid or freemium and the content has not been viewed within the last 24 hours. Saved in the API folder.  Circle "button" divs were created and activeModalStyling function tracks the stage "button" that should be highlighted in rattata.
5. Used the useEffect to move through the modals with a key press.
6. all img files were produced by grace in Figma and imported as svgs.
7. Removed Testimonals from the TestimonialsSect file so that the list could be sliced and exported into my component and it was cleaner.
CONCERNS and NEXT STEPS:
1. Do we want to have a "trigger" that would open the dialog box again upon clicking?  This is pretty easy to do.  We would just have to decide label and placement.
2. Mobile/regular tailwind classes -- I would love suggestions on how to get the styling more precise in the individual components, as it was difficult for me to do the styling inside of the Dialog component. Also, need comments regarding the content that should be included in the mobile view.  That was not in the figma sketches.
3. Let me know if the last_seen variable accomplishes the "viewing once in a session" goal or if there's a better approach.


LuckyHariharan:
Amazing PR! Looks like an amazing feature you're working on!!

components/studentPortal/freemium/FreemiumDialogComponent.tsx
```diff
@@ -0,0 +1,105 @@
+import { XIcon } from "@heroicons/react/outline";
+import { Close, Content, Overlay, Portal, Root } from "@radix-ui/react-dialog";
+import React, { useEffect, useState } from "react";
+import { useSelector } from "react-redux";
+import { getModalContent } from "../../../pages/api/studentPortal/freemium/getModalContent";
+import { themeSelector } from "../../../redux/themeSlice";
+import TrailAnimation from "../../ui/TrailAnimation";
+
+export enum ModalStage {
+  ONE = 1,
```
iamnotreddy, line 10: this can just be ModalStage = {ONE, TWO, THREE...}, enums have indexes automatically... not sure what you're doing here

kwiedenhaupt: cleaned this up.

components/studentPortal/freemium/FreemiumDialogComponent.tsx
```diff
@@ -0,0 +1,105 @@
...
+    if (currentStage === activeModal) {
+      styling = styling + " bg-rattata";
+    } else {
+      styling = styling + " bg-gray-300";
+    }
+    return styling;
+  };
+
+  useEffect(() => {
+    const handleKeyPress = (event: KeyboardEvent) => {
```
iamnotreddy, line 33: i think there's some opportunity to clean this function up, or at the least add some comments

pages/api/studentPortal/freemium/showModal.ts
```diff
@@ -0,0 +1,24 @@
...
+    const { userRole, userProfileData } = useSelector(profileSelector);
+    const lastSeenTime = userProfileData.lastSeen;
+    const lastSeenDateTime = new Date(lastSeenTime);
+    const currentDateTime = new Date();
+    const timeDifferenceInHours = differenceInHours(
+      currentDateTime,
+      lastSeenDateTime
+    );
+    if (
+      timeDifferenceInHours >= 24 &&
```
iamnotreddy, line 17: this is pretty clever, I think it should work. that being said I didn't see the modal immediately and I don't think I logged into skillify over the weekend 

one thought - can you make sure the timezones for the timestamp in the client and the database line up. want to make sure one is not being store in UTC and another in local

components/studentPortal/freemium/FreemiumDialogComponent.tsx
```diff
@@ -0,0 +1,105 @@
...
+              (stage) =>
+                currentStage === stage && (
+                  <TrailAnimation open={true}>
+                    {getModalContent(stage)}
+                  </TrailAnimation>
+                )
+            )}
+            <div className="flex flex-row justify-center space-x-2 absolute bottom-4 md:bottom-8 inset-x-0">
+              {[
+                ModalStage.ONE,
```
iamnotreddy, line 82: is there a cleaner way to loop through this array enums? if not, might be cleaner to declare this above in a variable then reference it down here

iamnotreddy (changes requested):
I think generally this looks pretty good. I think we'll need to refine the spacing, styling etc in the actual modals themselves, but we can perfect in future iterations. One thing we should fix though: looks like the content and circles are overlapping on a lot of modals.
Screen Shot 2023-03-20 at 11 45 34 AM

A general comment: I found some of the code a little hard to follow. I added a couple comments on opportunities for cleanup. If you can, I'd try to reduce the amount of logic that's in your JSX. 

Also, I think a good habit to get into (and one I can get more into myself!) is if a block of code isn't obvious, add a comment to explain.

components/studentPortal/freemium/FreemiumDialogComponent.tsx
kwiedenhaupt: I added comments, am open to cleaning it up more as well.

components/studentPortal/freemium/FreemiumDialogComponent.tsx
kwiedenhaupt: yes.  added the modalStageMappedArray, which maps through the enums and used that to map through displaying the components.

This cleans up the return and will allow for enum stages to be added, subtracted without having to change code in additional places.

pages/api/studentPortal/freemium/showModal.ts
kwiedenhaupt: I see your userRole is changed to "freemium" so it should work.
As far as I can tell, they are both in UTC time --
image
And I did use a function to convert them to UTC time and the output remained the same...I'll keep looking into this.

pages/api/studentPortal/freemium/showModal.ts
```diff
@@ -0,0 +1,25 @@
...
+    const lastSeenTime = userProfileData.lastSeen;
+    const lastSeenDateTime = new Date(lastSeenTime);
+    const currentDateTime = new Date();
+    const timeDifferenceInHours = differenceInHours(
+      currentDateTime,
+      lastSeenDateTime,
+    );
+
+    if (
+      timeDifferenceInHours <= 24 &&
```
vithushan19, line 18: I feel like this won't do what you want it to do. Last seen is updated when you load the student portal

kwiedenhaupt: the goal is to have the modal not come up too often.
I thought if they hadn't loaded the student portal in 24 hours that would  work...and then it would come up.  if they keep logged in and working, they probably won't keep needing the message.

Do you have any ideas or should I create something?

kwiedenhaupt:
image
ran test suite on showModal function

pages/api/studentPortal/freemium/showModal.ts
kwiedenhaupt: Would one solution be to add a column to the users table called lastLogin and in the SignInPage.tsx create a useMutation that writes a time back to the database when the user logs in and then use that value in my showModal function to see if it's been more than 24 hours?  
Another option would be to put the useMutation to write the time back to the database inside the showModal and label it lastSeenModal and then use that time in my showModal function to see if it's been more than 24 hours.

pages/api/studentPortal/freemium/showModal.ts
iamnotreddy: ah that makes sense...

@vithushan19 - would it be overkill to create an extra field in one of the hasura tables to keep track of this? i.e when they login and see the modal, we write to hasura that they saw it. And we read from hasura to decide whether to show the modal

not sure if there are better ways to do this

pages/api/studentPortal/freemium/showModal.ts
```diff
@@ -1,25 +1,32 @@
...
-  }
\ No newline at end of file
+  if (
+    timeDifferenceInHours !== null &&
+    timeDifferenceInHours >= 24 &&
+    (userRole === "freemium" || userRole === "paid")
+  ) {
+    dispatch(setLastSeenModal(currentDateTime))
+  updateLastSeenModal({ variables: { userId: user.uid, lastSeenModal: currentDateTime } });
+    return true;
```
iamnotreddy, line 26: you're telling redux and the database that the modal was seen before your function returns true (i.e shows the modal)... this might be the culprit behind the infinite re-rendering

pages/api/studentPortal/freemium/showModal.ts
```diff
@@ -1,25 +1,32 @@
...
+  } else {
+    dispatch(setLastSeenModal(currentDateTime))
+  updateLastSeenModal({ variables: { userId: user.uid, lastSeenModal: currentDateTime } });
```
iamnotreddy, line 29: the false case here is that the user didn't see the modal but you're still recording in the table that they saw it... you shouldn't do anything in this case

iamnotreddy (changes requested):
sorry for the delay on looking at this!

I confirmed that you're writing back to the hasura table correctly, but I was never able to see the modal even if I deleted myself from the freemium table. I logged the value of "showModal()" at the beginning of the studentPortal index component and this is what I see: 
image

The way you wrote the logic is causing an infinite render. Left a couple comments below on a potential trouble areas. You need to fix the process of getting the last time a user saw the modal, whether or not to show the modal, and telling the DB the user has now seen the modal. It's out of sync right now. 

Were you using redux intentionally? I'm not sure if you really need it and I think it's complicating your code. I'd consider dropping it and just using useQuery

vithushan19:
Can you sync with Grace on the mobile designs?

pages/api/studentPortal/freemium/showModal.ts
vithushan19: i dont' think it's overkill, you can also create a separate table and use a foreign key into the user table

pages/api/studentPortal/freemium/showModal.ts
vithushan19: I think we should create a new table and have a foreign key that links to the user table

vithushan19: and create a useMutation like you said

graphql/studentPortal/freemium/fetchLastSeenModal.ts
```diff
@@ -0,0 +1,21 @@
+import { gql } from "@apollo/client";
+
+export const FETCH_LAST_SEEN_MODAL = gql`
+query fetchLastSeenModal($_id: String = "") {
```
vithushan19, line 4: change this query to read from the freemium_user table

pages/studentPortal/index.tsx
```diff
@@ -65,6 +72,37 @@ export default function StudentPortalPage() {
     });
   }, [user]);
 
+  const [updateLastSeenModal] = useMutation(UPSERT_LAST_SEEN_MODAL);
+  const [showModal, setShowModal] = useState(false);
+
+  useQuery<FetchModalData>(FETCH_LAST_SEEN_MODAL, {
+    variables: {
+      _id: user.uid,
```
vithushan19, line 80: use the skip option along with the current users role to conditionally execute this query: https://stackoverflow.com/questions/59489097/use-query-hook-react-apollo-conditional-call

pages/studentPortal/index.tsx
```diff
@@ -65,6 +72,37 @@ export default function StudentPortalPage() {
...
+
+    onCompleted: (data) => {
+      if (data.users[0].freemium_user) {
+        const lastSeenValue = data.users[0].freemium_user?.lastSeenModal;
+        const lastSeenDifference = differenceInSeconds(
+          new Date(),
+          new Date(lastSeenValue)
+        );
+
+        if (lastSeenDifference > 60) {
```
vithushan19, line 91: will this show a user the freemium modal every minute?

pages/studentPortal/index.tsx
```diff
@@ -65,6 +72,37 @@ export default function StudentPortalPage() {
...
+          setShowModal(true);
+          updateLastSeenModal({
+            variables: { userId: user.uid, lastSeenModal: new Date() },
+          });
+        }
+      } else {
+        setShowModal(true);
```
vithushan19, line 98: is there a way not to repeat yourself with this else condition. this may not be possible, but see if it becomes easier once you are querying from the freemium_users table instead of the users table.

graphql/studentPortal/users/fetchUserRole.ts
```diff
@@ -9,13 +9,15 @@ query FetchUserRole($_id: String = "") {
     }
   }
 }
-
 `;
 
 export type FetchRoleData = {
-  users: UserData;
+  users: UserData[];
```
vithushan19, line 15: is this intentional?

vithushan19 (commented):
this looks great to me, especially the work around the saving to hasura and showing a dialog. I think we need to think through the onboarding flow more. Can we merge this PR in with just a basic radix dialog to start, and we can think through a more complex onboarding flow afterwards.

graphql/studentPortal/users/fetchUserRole.ts
kwiedenhaupt: removed.

pages/studentPortal/index.tsx
kwiedenhaupt: yes.  figured it out.

pages/studentPortal/index.tsx
kwiedenhaupt: I adjusted to 24 hours.

pages/studentPortal/index.tsx
kwiedenhaupt: great idea.  it works.

kwiedenhaupt:
I put the material in this in other PRs and no longer need this for reference.

PR #919, opened 2023-03-21
Typos and config

iamnotreddy:
Couple changes in here let me start with the small ones:
- changed the jsx property in tsconfig from "preserve" to "react"... this fixes the issue where the `import React from 'react'` disappears when you save
- fixed a couple typos i saw while scrolling through

For @gracejyu and @vithushan19:
I was scrolling through the site and thought the "What You Get" section was a bit hard to digest, so played around with a bulleted list. Here is the UI: 
Screen Shot 2023-03-21 at 12 20 13 PM

I think bullets for some of the sections is easier to digest. I also think there's some opportunity to update the content here overall, make it align with the new modals etc.... for example, I wonder if we can reword "success path coaching" because that term isn't immediately clear to me


components/landingPage/WhatYouGet.tsx
```diff
@@ -105,11 +113,12 @@ function FeaturesSection({
...
       <div className="flex flex-col w-full p-4 m-4 text-white max-w-7xl bg-murkrow rounded-xl">
-        {features.map((item) => (
-          <div className="flex flex-col m-4">
+        {features.map((item, index) => (
+          <div key={index} className="flex flex-col m-4">
             <p className="text-xl font-bold ">{item.title}</p>
             <p className="mb-2 text-xl text-charmander">({item.value})</p>
-            <p className="mb-2 text-xl ">{item.description}</p>
+            <Description description={item.description} />
+            {/* <p className="mb-2 text-xl ">{item.description}</p> */}
```
vithushan19, line 121: remove comment

PR #920, opened 2023-03-23
Setup Insert Mutation

LuckyHariharan:
In this PR I:

- setup the GraphQL insert mutation
- Implemented the useMutation hook when the user finishes the quiz (reaches the final stage)
- Setup an example object to pass to the database, since @jimip85 is developing the datastructure to collect all responses in another PR


LuckyHariharan:
@vithushan19 wanted to merge this in but I understand if this is too small for a PR. waiting for a PR from Jimi to merge in

vithushan19 (approved):
this looks great! small PRs are the best! this was super easy to review. I understand exactly what this change is going to do and how it will affect the bigger application. Nice work!

PR #925, opened 2023-03-28
added Freemium Info Icon on SideBar and locked components

kwiedenhaupt:
FreemiumSideBarItem.tsx
image

Web Development course locked feature
image

Goal of the PR: 
To create an info icon that we will use as a trigger for the modal that will welcome freemium/paid users to Skillify.

Main changes:
1. Added info icon
2. created FreemiumSideBarItem.tsx and pulled the disabling of the links/graying out content/showing toolbar into this component.  
3. placed FreemiumSideBarItem.tsx within the SideBarItem.tsx  and conditionally rendered it based off of the isGrayedOut value which is populated in the Sidebar based on a userRole query.
4. modified the TooltipComponent to take in an optional icon.
5. modified the SkeletonNodeView and the FreemiumSideBarItem to add the infoIcon.svg from Grace.

Questions/Thoughts:
1. Do we need to create a hover version of this icon that would be in another color like charmander so that upon hover the color is different? 
2. Do we need to modify the grayed-out sidebar items for freemium/paid users to not turn the charmander color on highlight -- in other words, should I take off the hover color and leave them gray for freemium/paid users?

Future Steps:
After the Modal PR is merged, the next step would be to link that modal to the info icon in the tooltip.


components/ui/TooltipComponent.tsx
```diff
@@ -27,7 +29,10 @@ const TooltipComponent: React.FC<TooltipComponentProps> = ({
             align="center"
             style={{ whiteSpace: "normal", width: "200px", lineHeight: "1.5" }}
           >
-            {message}
+            <div className="flex flex-row items-center">
+              {message}
+              <img src={icon} className="w-10 h-10 rounded-full p-2" />
```
vithushan19, line 34: should we use conditional rendering to only render the img element when the icon prop exists?

components/studentPortal/layout/SidebarItem.tsx
```diff
@@ -25,25 +26,33 @@ const SidebarItem = ({
...
-          }`}
-        >
-          <div>
-            {notifications ? (
-              <div className="relative left-6 top-1 ">
-                <div className="flex w-2 h-2 bg-red-500 rounded-full"></div>
-              </div>
-            ) : null}
-            {icon}
+      {isDisabled ? (
```
vithushan19, line 29: what was isDisabled being used for before? Are there any use cases where we'd want to disable a sidebar item but not sure the freemium tooltip?

I'm not sure that freemium sidebar item should live inside SidebarItem. maybe we should move this up one layer into the sidebar component? 

That way freemiumSidebarItem and SidebarItem don't know about each other and it is up to the Sidebar to organize them and determine which to show.

vithushan19 (commented):
this looks good, but I think we should move the freemium sidebar item up one level. let me know your thoughts.

components/ui/TooltipComponent.tsx
kwiedenhaupt: did icon &&.  thank you!

components/studentPortal/layout/SidebarItem.tsx
kwiedenhaupt: The isDisabled was created for the freemium users -- but your point is well-taken.  
I have moved the FreemiumSidebarItem to be conditionally rendered based on userRole === "freemium" || userRole === "paid" and moved it to the Sidebar, as it has now been refactored.

kwiedenhaupt:
I merged the new changes in the Sidebar into this branch and moved the FreemiumSidebarItem to the Sidebar.

I also added a pulsating animation to the info icon so that it is obvious it is clickable.  Let me know what you think of the changes!

vithushan19 (approved):
looks great! thank you for this

PR #926, opened 2023-03-28
Optimized LangQuiz w/ Enhanced bi-direc selection & response mgmt.

jimip85:
In this pull request, we have made some significant modifications to the following components:

**FirstProgrammingLanguageQuiz:**
(a) We've completely refactored the routing logic to be more reusable and efficient. The elimination of any hardcoded logic makes the quiz selection process more streamlined and faster.
(b) The quiz selection data structure has been modified into an object, making it easier to navigate through the quiz selection flow.

**SkillSelections:**
(c) We've refactored the SkillSelections component to improve the handling of selections and responses. Specifically, we've updated the integration of the handleStage function to ensure accurate quiz responses. By updating them with selected items for the current stage before resetting the current selection, we're making the quiz experience smoother and more efficient.

These changes will not only enhance the performance of the components, but also improve the overall user experience.


components/resources/quizzes/shared/SkillSelections.tsx
```diff
@@ -13,20 +18,49 @@ type SkillSelectionsProps = {
...
+  setQuizResponses,
+  currentStage,
   onNextClick,
   onBackClick,
   progress,
   title,
   body,
 }) => {
-  const [selected, setSelected] = useState<string[]>([]);
+  const currentSelections = selected[currentStage] || [];
```
LuckyHariharan, line 31: Nice!

components/resources/quizzes/shared/SkillSelections.tsx
```diff
@@ -13,20 +18,49 @@ type SkillSelectionsProps = {
...
+
   const handleSelection = (selection: string) => {
-    if (selected.includes(selection)) {
-      setSelected(selected.filter((s) => s !== selection));
+    const selectionIndex = currentSelections.indexOf(selection);
```
LuckyHariharan, line 34: what does indexOf do that map or filter can't? Asking because ive not used it and am curious.

pages/studentPortal/quizcomponent/FirstProgrammingLanguageQuiz.tsx
```diff
@@ -0,0 +1,183 @@
...
+  LEARNCODING,
+  WORKFIELDS,
+  BUILDTARGET,
+  BLUEPRINT,
+  RESULTS,
+}
+
+type QuizSelection = {
+  name: string;
+  result?: string;
```
LuckyHariharan, line 17: nice! good job on the type safety

pages/studentPortal/quizcomponent/FirstProgrammingLanguageQuiz.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const [stage, setStage] = useState(Stage.START);
+  const handleNextClick = () => {
+    setStage((prevStage) => prevStage + 1);
+  };
+  const handleBackClick = () => {
+    setStage((prevStage) => {
+      const newStage = prevStage - 1;
+      setQuizResponses((prev) => {
+        const resetResponseIndex = [...prev];
+        console.log(`this is the value of stage: ${newStage}`);
```
LuckyHariharan, line 102: remove this

pages/studentPortal/quizcomponent/FirstProgrammingLanguageQuiz.tsx
```diff
@@ -0,0 +1,183 @@
...
+    HTMLCSS: 0,
+    Python: 0,
+  });
+
+  const [stage, setStage] = useState(Stage.START);
+  const handleNextClick = () => {
+    setStage((prevStage) => prevStage + 1);
+  };
+  const handleBackClick = () => {
+    setStage((prevStage) => {
```
LuckyHariharan, line 98: good work!

LuckyHariharan (commented):
Great work Jimi! So many amazing changes in this PR. Let's communicate these results to Vithushan and get his review!

components/resources/quizzes/shared/SkillSelections.tsx
```diff
@@ -37,22 +71,43 @@ const SkillSelections: React.FC<SkillSelectionsProps> = ({
...
-                  : "bg-white text-black-600 border-2 border-black-500 rounded-xl"
-              }`}
-              onClick={() => handleSelection(selection)}
-            >
-              {selections[index]}
-            </div>
-          ))}
+          {selections.map((selection, index) => {
+            const isSelected = currentSelections.includes(selection);
+            const className = [
```
iamnotreddy, line 76: Easier and cleaner if you store the tailwind classes in a string. Something like this:

const baseClass = "list of classes here"

selectedClass = isSelected ? 'bg-violet-400' + baseClass : 'bg... + baseClass'

You can also use a string literal

components/resources/quizzes/shared/SkillSelections.tsx
```diff
@@ -37,22 +71,43 @@ const SkillSelections: React.FC<SkillSelectionsProps> = ({
...
+              "border-2",
+              "border-black-500",
+              "rounded-xl",
+              isSelected ? "bg-violet-300" : "bg-white",
+            ].join(" ");
+            return (
+              <div
+                key={index}
+                className={className}
+                onClick={() => handleSelection(selection)}
```
iamnotreddy, line 95: You can just reference the handle function directly without the parentheses or arrow syntax. Couple other cases of this below

components/resources/quizzes/shared/SkillSelections.tsx
```diff
@@ -37,22 +71,43 @@ const SkillSelections: React.FC<SkillSelectionsProps> = ({
...
+              </div>
+            );
+          })}
         </div>
         <div className="mt-4">
-          <Button label="Next" onClick={onNextClick} backgroundColor="yellow" />
+          <Button
+            label="Next"
+            onClick={() => {
+              handleStage();
```
iamnotreddy, line 106: You can't use shorthand syntax here when calling two functions so this is correct

pages/studentPortal/quizcomponent/FirstProgrammingLanguageQuiz.tsx
```diff
@@ -0,0 +1,182 @@
...
+}
+
+type QuizSelection = {
+  name: string;
+  result?: string;
+  weight?: number;
+};
+
+const FirstProgrammingLanguageQuiz = () => {
+  const quiz: QuizSelection[][] = [
```
iamnotreddy, line 22: This data structure is pretty confusing. Right now I read this as an array of an array of QuizSelection objects. 

Each element in the outer array is an array, but each inner array has only one element. So it's kind of pointless.

I think what you're going for is an array of objects with type QuizSelection, which makes sense to me. But you'll need to change your setter functions to reflect this if you change the data structure

tsconfig.json
```diff
@@ -1,7 +1,11 @@
 {
   "compilerOptions": {
     "target": "es5",
-    "lib": ["dom", "dom.iterable", "esnext"],
```
iamnotreddy, line 4: Remove this file from your PR we shouldn't be making changes to the typescript config file

pages/studentPortal/quizcomponent/FirstProgrammingLanguageQuiz.tsx
```diff
@@ -0,0 +1,182 @@
...
+    [
+      { name: "Learn an in-demand skill" },
```
iamnotreddy, line 24: Question for @LuckyHariharan as well

Are the quiz questions going to live in hasura? If so we should eventually fetch this from the db. If not this should move server side

iamnotreddy (commented):
I didn't get a chance to run this locally but nice job getting this to work.

That being said, lot of opportunity to improve this code - feel like we can simplify lot of things. Couldn't leave a full review since I'm on my phone, but I left a few comments that might help.

components/resources/quizzes/shared/SkillSelections.tsx
```diff
@@ -13,20 +18,49 @@ type SkillSelectionsProps = {
...
     }
   };
+
+  const handleStage = () => {
+    setQuizResponses((prev) => {
+      const updatedResponses = [...prev];
+      updatedResponses[currentStage - 1] = selected[currentStage];
+      return updatedResponses;
+    });
+    setSelected((prev) => {
```
vithushan19, line 58: does this even change the state at all?

pages/studentPortal/quizcomponent/FirstProgrammingLanguageQuiz.tsx
vithushan19: lets move to server side

vithushan19 (commented):
looking much better Jimi, nice work. there's still a few things i think that can be cleaned up.

components/resources/quizzes/shared/SkillSelections.tsx
jimip85: I was more interested in index #. Hence - I went with this approach.

components/resources/quizzes/shared/SkillSelections.tsx
jimip85: Thanks - changes have been implemented!

pages/studentPortal/quizcomponent/FirstProgrammingLanguageQuiz.tsx
jimip85: Good to know. I will have a conversation with Lucky to see next steps!

components/resources/quizzes/shared/SkillSelections.tsx
jimip85: I wrote it the way above to avoid this error: Type 'void' is not assignable to type 'MouseEventHandler<HTMLDivElement>'

components/resources/quizzes/shared/SkillSelections.tsx
jimip85: Yeah, this is how I push the selections to the parent component. If that's what you are asking... Otherwise, I'm not sure??

vithushan19: but you're just taking the previous state, and assigning it into a new variable and then re-assigning it to the same to the same state? 

isn't      `const updatedSelections = [...prev];`
the same thing as `const updatedSelections = prev` ?

pages/studentPortal/quizcomponent/FirstProgrammingLanguageQuiz.tsx
vithushan19: take a look at `pages/api/studentPortal/units.ts` I think it's easy enough to move these quiz question objects somewhere into the api folder. It'll clean up this component, and separate the UI from the logic a bit more. 

I think it's a small enough change to do in this PR

pages/studentPortal/quizcomponent/FirstProgrammingLanguageQuiz.tsx
```diff
@@ -0,0 +1,182 @@
...
+      return options.name;
+    });
+  });
+
+  type PageContent = {
+    title: string;
+    body: string;
+  };
+
+  const quizQuestions: PageContent[] = [
```
vithushan19, line 65: This is already in a much better place since when you initially started, so great work on cleaning up this code already.

I want to point out some pieces that are confusing.

You have two variables called quiz and quizQuestions, I can't tell which one is supposed to be the single source of truth for quiz state, and how the variables are related. 

The names of your variables don't really match the name of the type either.

What if you created a type called `QuizData` and all the quiz questions and all the title and body texts were captured inside of that object. 

That QuizData object would be the single source of truth for the quiz. You're basically combining the quiz and quizQuestions into a single variable.

I would differentiate between QuizData and QuizState. QuizData just has the info about the quiz questions and weight/title/body text. QuizData has no idea about what the user has selected.

QuizState would have the quiz data inside of it, but also hold information about what the user has currently selected.

I'd also be careful to differentiate between a quiz option and a quiz selection.

pages/studentPortal/quizcomponent/FirstProgrammingLanguageQuiz.tsx
```diff
@@ -0,0 +1,182 @@
...
+      body: "Select 1-3 choices.",
+    },
+    {
+      title: "What are you interested in building?",
+      body: "Select all that apply.",
+    },
+  ];
+
+  const [quizResponses, setQuizResponses] = useState([]);
+  const [selectedSkill, setSelectedSkill] = useState<Array<Array<string>>>([]);
```
vithushan19, line 85: what is the difference between quizResponses, and selectedSkill?

I think if you change Array<Array<string>> to a custom QuizState object that might clear things up, but I'm not sure how your two states are different

pages/studentPortal/quizcomponent/FirstProgrammingLanguageQuiz.tsx
```diff
@@ -0,0 +1,182 @@
...
+      const newStage = prevStage - 1;
+      setQuizResponses((prev) => {
+        const resetResponseIndex = [...prev];
+        return resetResponseIndex;
+      });
+      return newStage;
+    });
+  };
+
+  const renderStage = () => {
```
vithushan19, line 108: this part is super clean, it's just the state management that needs improvement

pages/studentPortal/quizcomponent/FirstProgrammingLanguageQuiz.tsx
vithushan19: I might be missing something, but I think you can clean up these state variables and come up with a cleaner/more readable solution

vithushan19 (commented):
getting better, but still room for improvement especially around cleaning up state.

pages/studentPortal/quizcomponent/FirstProgrammingLanguageQuiz.tsx
jimip85: Resolved!

pages/studentPortal/quizcomponent/FirstProgrammingLanguageQuiz.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+  const [stage, setStage] = useState(Stage.START);
+  const handleNextClick = () => {
+    setStage((prevStage) => prevStage + 1);
+  };
+  const handleBackClick = () => {
+    setStage((prevStage) => {
+      const newStage = prevStage - 1;
+      setQuizResponses((prev) => {
+        const resetResponseIndex = [...prev];
```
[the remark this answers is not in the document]
jimip85: I removed this function as a whole!
- After my optimization to the skillSelection component. I no longer required this.

components/resources/quizzes/shared/SkillSelections.tsx
```diff
@@ -13,22 +17,37 @@ type SkillSelectionsProps = {
...
   progress,
   title,
   body,
 }) => {
-  const [selected, setSelected] = useState<string[]>([]);
+  const currentSelections = selected[currentStage] || [];
   const handleSelection = (selection: string) => {
-    if (selected.includes(selection)) {
-      setSelected(selected.filter((s) => s !== selection));
+    const selectionIndex = currentSelections.indexOf(selection);
```
vithushan19, line 31: what happens if a quiz has the same string on two different questions?

vithushan19 (changes requested):
- move data/logic to the pages/api folder
- create a new type called `QuizData`
- - stores info about the quiz questions (ie. options, title, body)
- - remove the PageContent type in favour of this new type
- create a new type called `QuizState`
- - has a property called quizData: QuizData
- - also stores info about the current selections

components/resources/quizzes/shared/SkillSelections.tsx
jimip85: That works fine!

pages/api/studentPortal/quizzes/FirstLang.ts
```diff
@@ -0,0 +1,67 @@
+type QuizData = {
+    name?: string;
```
vithushan19, line 2: this is not good, because all of the properties in this type are optional. You're kind of using this type as a placeholder, than can have any or all of these properties. 

Think about what an instance of a QuizData is conceptually inside of our business. Imagine having multiple quizzes, each with their own quiz data, what makes up a quiz? Think about the exercise we did with creating a design for different components in the system. You want your code to be a mirror of your mental model of quizzes.

pages/api/studentPortal/quizzes/FirstLang.ts
```diff
@@ -0,0 +1,67 @@
...
+  ];
+
+  export const selectionsPerPage = quizQuestions.map((pages) => {
+    return pages
+      .map((selections) => {
+        if (selections.name !== undefined && selections.name !== null) {
+          return selections.name;
+        }
+      })
+      .filter(Boolean);
```
vithushan19, line 66: what does this filter do?

pages/api/studentPortal/quizzes/FirstLang.ts
```diff
@@ -0,0 +1,67 @@
...
+      { name: "Email", result: "HTML/CSS", weight: 1 },
+      { name: "Tools to automate your life", result: "JavaScript", weight: 1 },
+      { name: "I don't know / Not sure yet", result: "JavaScript", weight: 1 },
+    ],
+  ];
+
+  export const selectionsPerPage = quizQuestions.map((pages) => {
+    return pages
+      .map((selections) => {
+        if (selections.name !== undefined && selections.name !== null) {
```
vithushan19, line 62: this is not a clean way of getting selections. you have map through your array and look for all the objects that have a name, and you are assuming that is what you're looking for. your code is very prone to causing bugs or errors

pages/resources/quizzes/FirstLang/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+            title={quizQuestions[stage][stage].title}
+            body={quizQuestions[stage][stage].body}
+          />
+        );
+      case Stage.LEARNCODING:
+        return (
+          <SkillSelections
+            onNextClick={handleNextClick}
+            onBackClick={handleBackClick}
+            title={quizQuestions[stage][stage - 1].title}
```
vithushan19, line 47: this is not great either, you are assuming that the title you are looking for will always be at the top of the array, which isn't guaranteed. how will you prevent a new team member from accidentally adding a question to the top of the array?

vithushan19 (commented):
I think you're really close, you just need to use something else instead of a 2D array. You need a custom type that accurately represents your mental model of the quizzes.

pages/api/studentPortal/quizzes/FirstLang.ts
vithushan19: what if you used this for the QuizData type: 

```
type QuizQuestion = {
  title: string;
  body: string;
  options: QuizOption[];
  maxSelections?: number;
};

type QuizOption = {
  name: string;
  result?: string;
  weight?: number;
};

type QuizData = {
  title: string;
  body: string;
  questions: QuizQuestion[];
};
```

pages/api/studentPortal/quizzes/FirstLang.ts
```diff
@@ -0,0 +1,67 @@
+type QuizData = {
```
vithushan19, line 1: rename this file to firstProgrammingLanguage.ts only components and pages are capitalized

pages/api/studentPortal/quizzes/firstProgrammingLanguage.ts
```diff
@@ -0,0 +1,80 @@
+export type QuizQuestion = {
```
vithushan19, line 1: this file is called firstProgrammingLanguage, so everything in this file should be specific to that quiz. These custom types can actually be moved into a shared file. maybe we should make a file called `types.ts` and move these shared custom types into that file.

pages/resources/quizzes/FirstLang/FirstProgrammingLanguageQuiz.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+const FirstProgrammingLanguageQuiz = () => {
+  const [stage, setStage] = useState(Stage.START);
+  const [quizViewState, setQuizViewState] = useState<QuizViewState>();
+
+  useEffect(() => {
+    const quizViewState = {
+      title: quizData.title,
+      body: quizData.body,
+      questions: [],
```
vithushan19, line 26: instead of initializing this prop with an empty array and pushing questions into this array you could use the map function to clean this up and remove the for loop.

ie. 

```
questions: quizData.questions.map(question => {
  return {
    title: question.title,
    body: question.body,
    options: question.options,
  }
}
```

great job figuring out the for loop approach, but you need to start thinking about data as functional instead of imperative. Imperative programming is very junior. You think of code as being executed one line at a time.

To get yourself to the next level, you need to think of arrays as a stream of data that can be filtered or mapped. This is a more functional way of thinking about code.

pages/resources/quizzes/FirstLang/FirstProgrammingLanguageQuiz.tsx
```diff
@@ -0,0 +1,116 @@
...
+    setQuizViewState(quizViewState);
+  }, []);
+
+  const handleNextClick = () => {
+    if (
+      stage == Stage.QUESTIONS &&
+      quizViewState.currentQuestion < quizData.questions.length - 1
+    ) {
+      setQuizViewState({
+        ...quizViewState,
```
vithushan19, line 53: love this! great use of spread

pages/resources/quizzes/FirstLang/FirstProgrammingLanguageQuiz.tsx
```diff
@@ -0,0 +1,116 @@
...
+    if (stage == Stage.QUESTIONS && quizViewState.currentQuestion > 0) {
+      setQuizViewState({
+        ...quizViewState,
+        currentQuestion: quizViewState.currentQuestion - 1,
+      });
+    } else setStage((prevStage) => prevStage - 1);
+  };
+
+  const handleOptionClick = (option: QuizOptionViewState) => {
+    // TODO create a new quizViewState and then call setQuizViewState
```
vithushan19, line 68: guessing this will come in a follow up PR?

pages/resources/quizzes/FirstLang/FirstProgrammingLanguageQuiz.tsx
```diff
@@ -0,0 +1,116 @@
...
+        );
+      case Stage.RESULTS:
+        return <LangResults onBackClick={handleBackClick} />;
+      default:
+        return null;
+    }
+  };
+  return (
+    <>
+      <div>{quizViewState?.currentQuestion}</div>
```
vithushan19, line 107: did you mean to include this line?

vithushan19 (approved):
Awesome work Jimi! You worked super hard on this PR, and I noticed your improvement over the course of this PR itself. The more you stick to this process, the more you are going to improve. You should be proud of your commitment and perserverance in continually addressing the feedback and being coachable to get this code in a really good place. 

It's gotten so much better compared to your first version. You did a great job working with Lucky and taking my feedback to get his ready to merge.

I left a few comments where you could improve the code, but nothing blocking.

pages/resources/quizzes/FirstLang/FirstProgrammingLanguageQuiz.tsx
vithushan19: @LuckyHariharan you should have noticed this as well.

pages/api/studentPortal/quizzes/firstProgrammingLanguage.ts
vithushan19: @LuckyHariharan make sure these types are moved before you use them in your career quiz

jimip85: The types.ts file lives in the quizzes/shared folder now.

pages/resources/quizzes/FirstLang/FirstProgrammingLanguageQuiz.tsx
jimip85: awesome! I didn't think of this!

pages/resources/quizzes/FirstLang/FirstProgrammingLanguageQuiz.tsx
jimip85: Yup! Didn't want to make the existing PR any larger.

PR #927, opened 2023-03-28
GraphQL mutation for FirstProLang (WIP)

jimip85:
This pull request adds the INSERT_CODING_QUIZ_RESPONSE mutation to the existing GraphQL schema. 

Here are the technical details:
- Utilizes Apollo Client's gql function to define the mutation
- Accepts an array of coding_quiz_insert_input objects as input
- Inserts the input objects into the 'coding_quiz' table
- Returns an array of inserted objects with the 'name', 'result', and 'email' fields

These changes will enable the client to insert coding quiz responses into the database with ease and retrieve the necessary data for analysis.

### DIFF
--- a/graphql/quizzes/insertFirstProLang.tsx
+++ b/graphql/quizzes/insertFirstProLang.tsx
@@ -1,0 +1,13 @@
+import { gql } from "@apollo/client";
+
+export const INSERT_CODING_QUIZ_RESPONSE = gql`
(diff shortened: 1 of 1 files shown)


LuckyHariharan (changes requested):
Great work on the mutation! I think it would be helpful to finish the rest of it, and send off example data to the database using this mutation, when the final stage of your quiz is reached. [this might be handy](https://www.apollographql.com/docs/react/data/mutations/ ).  Take a look at my career-quiz.tsx page to see how I used useMutation to do this.

PR #929, opened 2023-04-01
Cleaner Freemium Modal Component

kwiedenhaupt:
Purpose of this PR:
To create an onboarding experience for the Freemium User Journey.  This PR simply sets up the logic that moves between the three modals in a way that is dynamic so that modals can be added/deleted without having to refactor code.

Modal 1
image
Modal 2
image
Modal 3
image

INCLUDED IN THIS PR
1. a function called getModalContent is used to map the content to the stages.
2. for the first stage there is no back button, but an invisible one is rendered so spacing is kept exact
3. for middle stages both back and previous buttons are rendered
4. for the last stage a back button is rendered and a done non-clickable button is rendered to indicate there are no more stages.
5. I put <Close> tags around the done button so that it would close the modal as well.

NOTES:
Should I remove the gray x on the top right?  is it overkill?
Let me know if I should remove the comments


vithushan19 (approved):
this is awesome! great work!

PR #930, opened 2023-04-02
Kw freemium modal logic

kwiedenhaupt:
In this PR the goal was to create the logic surrounding the PR when to display the freemium onboarding modal.

STEPS:
1. create a fetchLastSeenModal query to determine if/when the modal was last seen
IN THE ONCOMPLETED:
2. logic to render the modal only if the query was seen more than 24 hours ago 
3. created an upsert query called updateLastSeenModal that will enter a value into the freemiumUsers table for the lastSeenModal with a new Date () value if it is null or overwrite it, if it exists
4. set showModal to true if the modal is shown -- this will be the trigger to render the modal in the return -- note that if the user is not freemium or paid, the query will be skipped altogther and this value will be false by default and the modal will not render.
OTHER STEPS:
5. using the userRole value that is saved into redux from the fetchUserRole query run in the Sidebar and then skipped running the query if users are not freemium or paid


vithushan19 (approved):
looks great!

vithushan19 (approved):

PR #934, opened 2023-04-08
Upgrade Modal

kwiedenhaupt:
PR Goal: 
Populate the second modal that will be used to direct users to upgrade to full version after their trial ends.

REGULAR VIEW: 
image
MOBILE: 
image

NOTES: 
I added a series of colors called freemium-200, 500, 900 into the tailwind config file.
These are used in the Figma sketches and are a series of blues from almost white to dark blue. 
2. create a white band at the bottom of the screen like the figma sketches that encapsulates the prev and next buttons to


tailwind.config.js
```diff
@@ -49,7 +49,11 @@ module.exports = {
           transparent: "#000000e1",
           500: "#000000",
         },
-        sky: "#E2E3FB"
+        freemium: {
```
vithushan19, line 52: I don't think these colours should be named after the feature. We would want the colours to be consistent across the app. If one day we wanted to make the skillify dark purple into a lighter shade, we'd want that change to work across the app, instead of having to modify the freemium feature indepdently. 

I don't even see you using freemium-900 in this PR.

and 500 looks close to our existing rattata colour. are you able to reuse that one? 



Are any of these colours the same as the murkrow color already defined in this config? If so, you can just reuse that colour. If not, you could add these numbered variants under the murkrow colour, but then you'll have to update everywhere murkrow is used.

freemium-900 looks identical to murkrow, so you probably just use that. 

let's create a new colour called `mewtwo` instead of freemium-200

and

vithushan19 (approved):
looks good, but your current implementation of the new colours is less maintainable over time.

tailwind.config.js
kwiedenhaupt: In looking closer at the rbg, I see my freemium-500 is the rattata, but with 20 percent opacity, so I was able to write that in tailwind easily! and you are right about bg-murkrow.  Thanks for pointing that out.
I added the mewtew as well. I was thinking about adding to the rattata, but had the same thought, that things would need to be changed throughout the app.  Thanks for explaining your reasoning on how/why you'd do this.

PR #936, opened 2023-04-09
Where to Start Modal

kwiedenhaupt:
GOAL OF PR: 3rd modal called "Where to Start" that will direct freemium users to the proper course during their onboarding experience.

Regular View
image

Mobile View
image

ADDITIONAL CHANGES:
Removed the "X" button in the FreemiumDialogueComponent that closes the modal.  The "Done" button closes the modal on the last modal and the "X" button doesn't look good and crowds the view on mobile.


tailwind.config.js
```diff
@@ -49,7 +49,7 @@ module.exports = {
           transparent: "#000000e1",
           500: "#000000",
         },
-        sky: "#E2E3FB"
+        mewtew: "#E2E3FB"
```
vithushan19, line 52: should be spelled mewtwo

PR #937, opened 2023-04-10
plans component

kwiedenhaupt:
Purpose of the PR:
To create a plans page that contains cards that outline plans that students can sign up for

Desktop View:
image

Mobile View:
image
image
image

Structure of the Plans Page
1. Index page that houses the PlansCard. The data is contained in an array of cards so that it can easily be expanded or retracted.
2. PlansCard contains a PlansRow component. Again, the data is contained in an array that is mapped out.
3. PlansRow component has the redX and greenCheck images taken from the Figma sketches and then a description next to it.  It is set up as a grid so that the description is able to populate the width of the card.

TO FOLLOW IN ANOTHER PR:
1. Populate the Data
2. Create the clickable buttons that contain the paths that we want to send them to with handleClick functions that are drilled down.


pages/studentPortal/plans/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+
+const Plans = ({ children }: PlansProps) => {
+  const plansCardData: PlansCardProps[] = [
+    {
+      title: "Card 1",
+      description: "desc 1",
+      price: "$0",
+      buttonLabel: "Sign Up",
+      onClick: () => console.log("clicked"),
```
iamnotreddy, line 17: remove the console.logs

pages/studentPortal/plans/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+      description: "desc 2",
+      price: "more",
+      buttonLabel: "Book a Call",
+      onClick: () => console.log("clicked"),
+    },
+  ];
+
+  return (
+    <div>
+      <div>
```
iamnotreddy, line 30: extra div

iamnotreddy (changes requested):
things to address:
- fix unnecessary spacing both on the main page and inside the card
- add a spacing on the left / right so the row PlanRow doesn't touch edge of the card
- let's try to get this render without a scroll-bar, maybe take a look at overflow in the tailwind docs
-

kwiedenhaupt:
CHANGES MADE: 
1. changed the rendering to getServerSideProps
2. re-structured the data so that the planRowData[] is included in the plansCardData[].
3. adjusted spacing of the cards
4. added content and links to the onClick properties of the buttons
5. changed the spacing so that no matter if there are varying numbers of columns the cards will start at the same height
6. added the ....and more! conditional optionalMessage

FURTHER CONCERN:
it seems like I am still having the overflow problem? I tried throwing the overflow-hidden class into the parent container of the plans page and it didn't seem to help...same with the plansCard.  There is no fixed height to the plans card that would be inadvertently causing it to be longer than I expected.  not sure.

kwiedenhaupt:
image

pages/plans.tsx
```diff
@@ -0,0 +1,131 @@
+export const plansCardData = [
```
iamnotreddy, line 1: let's declare this inside the getServerSieProps function instead of at the top of the file

iamnotreddy (approved):
couple small changes but nice work!

vithushan19 (approved):
lgtm

pages/plans.tsx
```diff
@@ -0,0 +1,131 @@
...
+      },
+    ],
+  },
+  {
+    title: "Premium",
+    description: "Contact us for pricing",
+    price: "Custom",
+    buttonLabel: "Book a Call",
+    onClick: "https://joinskillify.com/call",
+    optionalMessage: "...and more!",
```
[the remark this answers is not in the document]
kwiedenhaupt: done!

kwiedenhaupt:
Updated UI with the changes:
image

PR #938, opened 2023-04-11
14-day trial length

kwiedenhaupt:
Header on load:
image

This just included changing a hard-coded variable TOTAL_TRIAL_DAYS in two files.


iamnotreddy (approved):
If we change the trial length, we have to update this value in two separate places. I think this is ok but if the scope expands, we'll want to hold this value in one place.

vithushan19 (approved):

PR #939, opened 2023-04-11
exit modal - logic only

kwiedenhaupt:
Purpose of PR: 
To test the logic that will trigger the exit modal.  This does NOT contain the new Freemium Component that will need to be created.  It displays the onboarding modal for testing purposes.

NOTE: 
1.  do we want this in a useEffect? or what is best with that?
2. TOTAL_TRIAL_DAYS needs to be hard coded in three places the way this is set up overall in the repository.  is that okay or do you have a better suggestion?
3. when the trialdaysremaining is 0 the modal shows.  do we want them to have an extra day so they have access on day 0?
4. how do we then disable all the content????


iamnotreddy (changes requested):
Tested the logic on my user and it seems to work so no issues there. Left a couple quick comments.

Re: disabling all the content, played around a bit and one approach is to pass a boolean prop into your exit modal component, then set open to true. 

One thought for @vithushan19 - part of me thinks we store the exit modal display logic inside the onComplete of the useQuery that checks whether to show the onboarding modal. Seems more congruous to check for modal logic in one place instead of two. We could also consolidate some of the code into one function that checks which of the two modals to show since we would never show both. 

This does add more complexity to the onComplete (and maybe kinda violates separation of concerns?) but eliminates the need for an extra effect / keeps the studentPortal index page emptier

LuckyHariharan:
> Purpose of PR: To test the logic that will trigger the exit modal. This does NOT contain the new Freemium Component that will need to be created. It displays the onboarding modal for testing purposes.
> 
> NOTE:
> 
> 1. do we want this in a useEffect? or what is best with that?
> 2. TOTAL_TRIAL_DAYS needs to be hard coded in three places the way this is set up overall in the repository.  is that okay or do you have a better suggestion?
> 3. when the trialdaysremaining is 0 the modal shows.  do we want them to have an extra day so they have access on day 0?
> 4. how do we then disable all the content????


Finally took a look at this PR and its super cool! My 2 cents worth on points 1 and 3: 
1. I think a useEffect to trigger the exit modal. 3. Yes, I think it is better design to allow them day 0 -- to help introduce more students to the full paid skillify membership by giivng them some extra time.

kwiedenhaupt:
Major Changes Made to this PR Recently:
1.  elapsedDays and calculateRemainingTrialDays functions were updated.
2. helpers.test.ts was also modified to catch up with the changed functions
3. Moved the FreemiumExitModal to the Layout.tsx page making the user unable to click out of the modal no matter what page they try to navigate to.

PLEASE NOTE: 
The content in the FreemiumExitModal is contained in a following PR (946) that is branched off this PR.  I did this because I thought it would make it easier to review.

iamnotreddy:
@kwiedenhaupt - i felt like the modal logic was cluttering the layout component so I created a custom hook. I'm not sure if you've used this pattern before but it's useful for isolating logic and organizing code.

Basically I took all the relevant state and useQuery hook and moved it into it's own React functional component that we can call inside the layout component. The custom hook takes in userId, userRole, and createdAt and returns showOnboardingModal and showExitModal as well as updating when the user last saw the modal.

iamnotreddy (approved):
nice work on this

PR #944, opened 2023-04-14
create api route for components lesson; implement server side loading

iamnotreddy:
I tried to follow your implementation for the HTML lesson but it's not working - not sure if you finished it or maybe I'm missing something? 

I tried a slightly different approach - lmk if you think this is viable before I implement in other areas:
- I wanted to call the API route directly inside getServerSideProps but Next doesn't allow this (or discourages it) because it creates an extra request ([link](https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props#getserversideprops-or-api-routes))
- A [stackoverflow](https://stackoverflow.com/questions/65752932/internal-api-fetch-with-getserversideprops-next-js) workaround I found was to define the "fetching" logic in a separate function and reuse that in both the API route and getServerSideProps. This way one block of code is controls the lesson data.
- The naming is a little confusing in this test case because we refer to the general lesson data as LessonComponentData and this is the components lesson 😅 
- One downside I can see to this approach is creating a unique function name for each API route. Not sure if there's an easy way to solve for that


pages/api/studentPortal/web/react/components.ts
```diff
@@ -0,0 +1,97 @@
+import { NextApiRequest, NextApiResponse } from "next";
+import { LessonComponentData } from "../../../../../components/studentPortal/lessons/LessonComponent";
+
+export default async (req: NextApiRequest, res: NextApiResponse) => {
```
vithushan19, line 4: I would create a folder called reactjs inside the `pages/api/studentPortal/lessons` folder and move this file into there.

https://github.com/skillify-ca/skillify-web/tree/main/pages/api/studentPortal/lessons/basics

pages/studentPortal/web/React/components.tsx
```diff
@@ -1,21 +1,26 @@
 import { useRouter } from "next/router";
-import React from "react";
 import LessonComponent, {
   LessonComponentData,
 } from "../../../../components/studentPortal/lessons/LessonComponent";
 import { Button } from "../../../../components/ui/Button";
+import { getComponentsLessonData } from "../../../api/studentPortal/web/react/components";
+
+const React2 = (props: { lessonComponents: LessonComponentData[] }) => {
```
vithushan19, line 8: so if the api is working correctly, we should be able to delete this file in favour of using this dynamic component:

https://github.com/skillify-ca/skillify-web/blob/main/pages/studentPortal/intro/%5B...slug%5D.tsx#L65

Currently that dynamic lesson component is hard-coded to the basics course. We will need to use nextjs dynamic routing to support a dynamic course and lesson name.

https://nextjs.org/docs/routing/dynamic-routes

This file will have to be moved from `pages/studentPortal/intro/[...slug]` and to `pages/studentPortal/[courseId]/[lessonId]`

vithushan19 (commented):
you'll know you're on the right track if you're able to delete your `pages/studentPortal/web/React/components.tsx` file

vithushan19:
this is looking like it's on the right track, let me know if you need help with testing it

tsconfig.json
```diff
@@ -12,7 +12,7 @@
     "moduleResolution": "node",
     "resolveJsonModule": true,
     "isolatedModules": true,
-    "jsx": "react",
+    "jsx": "preserve",
```
vithushan19, line 15: don't commit this

pages/studentPortal/lessons/[...lesson].tsx
```diff
@@ -0,0 +1,22 @@
+/* eslint-disable react/react-in-jsx-scope */
+
+const Lesson = (props: { path }) => {
+  //
+  return (
+    <div>
+      <p className="text-2xl">HELLLOO</p>
+      <p>{JSON.stringify(props.path)}</p>
```
Comment (vithushan19) [inline pages/studentPortal/lessons/[...lesson].tsx:8]: should we delete this file?

pages/api/studentPortal/lessons/web/index.ts
```diff
@@ -0,0 +1,12 @@
+import { getLessonForReactCourse } from "./react";
+
+export function getCourseForWebSection(courseId: string, lessonId: string) {
+  if (courseId === "react") {
+    return getLessonForReactCourse(lessonId);
+  } else if (courseId === "graphql") {
+    // return getLessonForGraphQLCourse(lessonId) function once complete
```
vithushan19, line 7: maybe remove this else case if this lesson isn't ready or is unused

pages/studentPortal/lessons/web/[...slug].tsx
```diff
@@ -0,0 +1,43 @@
+import { useRouter } from "next/router";
+import LessonComponent, {
+  LessonComponentData,
+} from "../../../../components/studentPortal/lessons/LessonComponent";
+import { Button } from "../../../../components/ui/Button";
+import { getCourseForWebSection } from "../../../api/studentPortal/lessons/web";
+import LessonPage from "../../intro/[...slug]";
+
+const React2 = (props: { lessonComponents: LessonComponentData[] }) => {
```
Comment (vithushan19) [inline pages/studentPortal/lessons/web/[...slug].tsx:9]: we will eventually want to move towards just using this file for all web/React lessons. 

If you look at all your lesson components in this PR, they should all start to look very similar. 

In a follow up PR, you should delete all those react lesson components, and just use this file to display any react lessons. You could use the slug to figure out which method in your api you should call.

Do that in a new PR though, this one has been open long enough.

vithushan19 (approved):
looks good to me. Dont merge in the change to the tsconfig. 

After you merge this in, figure out how you can delete all the react lesson components and just a single slug page to show all react lesson. 

This PR is a good first step to creating the necessary endpoints to power our mobile apps. But even after this PR is merged, there is work to be done on connecting these new endpoints to the react client.

PR #945, opened 2023-04-19
Animations into FirstProgrammingQuiz

jimip85:
In this PR, 

- I integrated the FirstProgrammingQuiz routing component with Lucky's QuizTransition component, which animates the transition between each stage.

- To enhance the animation, I updated the configuration with a transform property that animates each item from left to right when triggerAnimation is true.

With these updates, the QuizTransition component provides a more visually appealing and seamless user experience by offering a smooth transition between stages.


iamnotreddy (approved):
nice work

vithushan19 (approved):
nice!

PR #946, opened 2023-04-21
Content for Exit Modal and modifications to FreemiumExitModal.tsx

kwiedenhaupt:
Purpose of this PR: 
I wanted to make it easier to review -- rather than adding it onto PR939, Exit Modal, I decided to branch off kw-exitModal and create a new branch, kw-exitModalContent, so that you could review solely the exit modal itself.

Regular View:
image
Mobile View:
image

Changes Made in this PR:
1. modified FreemiumExitModal to be only 1 modal and styled it accordingly.
2. populated the modal with the appropriate content.

Also note: Due to changes made on the PR839 Exit Modal, the exit modal now lives on the Layout.tsx page so that it always loads and the user can never navigate to a direct link and have access to any of the portal content if they are an expired user.


iamnotreddy (approved):
I noticed when the exit modal triggers and you click view plans, you actually can't un-toggle the modal. This is because the modal is triggered from the layout component and its open prop is always set to true.

I committed some code that solves for this. I followed the code we use on non studentPortal pages by defining a getLayout function on the Plans component. My understanding is this prevents the <Layout> component from wrapping around Plans (and therefore gets around the ExitModal). I think this makes sense.

I also added the NavBar to the top of this page, and I added a Plans section to the NavBar

components/studentPortal/freemium/modals/Exit.tsx
```diff
@@ -0,0 +1,21 @@
+import router from "next/router";
+import React from "react";
+import { Button } from "../../../ui/Button";
+
+const handlePlansSignUp = () => {
+  router.push("/plans");
+};
+
+export default function Exit() {
```
iamnotreddy, line 9: I'd probably name this component ExitModal because it could be confusing when you call "Exit" inside code

kwiedenhaupt:
@iamnotreddy that makes a lot of sense concerning not having the Layout component loaded so that it is not affected by the modal -- in fact, as you have indicated -- those pages that are not in the student portal we wouldn't want affected by the modal either.  so great fix.  and thanks for adding plans to the nav bar!

kwiedenhaupt:
@iamnotreddy does it make sense that I added the "Plans" to the navbar like "Quiz" and "Blog" so that it comes up for the blog as well?

PR #947, opened 2023-04-21
Connect Modal to the pulsating info icon in the tooltip

kwiedenhaupt:
Purpose of the PR:
When the user clicks on the tooltip icon that the freemium modal will appear

Tooltip when user hovers over the greyed-out items:
image
When the icon is clicked on: 
image

CHANGES MADE: 
1. Placed the FreemiumDialogueComponent (modal) inside of the tooltip.  
2. Made the icon that was pulsating a trigger to open the modal.
3. Created the handleModalToggle to keep track of whether the icon was clicked on and used that state to render the FreemiumDialogueComponent at the bottom.


iamnotreddy (changes requested):
mostly looks good! but two things I noticed: 
- when you click on the tooltip info icon, it doesn't navigate directly to the premium modal
- if you click out of the modal and try to click back in, you have to click twice to launch it again

kwiedenhaupt:
@iamnotreddy I added the handleModalClose function to reset the isOpenModal to false after the modal closes.  

In order to do that -- I put that in the onClose prop of the FreemiumDialogComponent and then added the prop onOpenChange that takes in the onClose prop to the Root component of the FreemiumDialogComponent to control the onClose functionality.

This should fix the problem of having to click the info icon twice to get the modal to open.  it was previously set to true from the first opening of the modal, and then when you closed the modal and wanted to open it again, you had to click it once to re-set it to false, and then again to true -- and that would then by the second time cause the modal to show.

kwiedenhaupt:
@iamnotreddy I'm confused about the modal not coming up right away when you click on the info icon in the tooltip.  it comes up right away for me.  any ideas?

iamnotreddy:
> @iamnotreddy I'm confused about the modal not coming up right away when you click on the info icon in the tooltip. it comes up right away for me. any ideas?

ah what I meant was having it directly navigate to this stage of the modal:

Screen Shot 2023-04-24 at 2 26 33 PM

iamnotreddy:
> @iamnotreddy I added the handleModalClose function to reset the isOpenModal to false after the modal closes.
> 
> In order to do that -- I put that in the onClose prop of the FreemiumDialogComponent and then added the prop onOpenChange that takes in the onClose prop to the Root component of the FreemiumDialogComponent to control the onClose functionality.
> 
> This should fix the problem of having to click the info icon twice to get the modal to open. it was previously set to true from the first opening of the modal, and then when you closed the modal and wanted to open it again, you had to click it once to re-set it to false, and then again to true -- and that would then by the second time cause the modal to show.

nice just tested this and it works!

components/studentPortal/freemium/FreemiumDialogueComponent.tsx
```diff
@@ -20,15 +20,19 @@ export interface FreemiumDialogComponentProps {
...
 const FreemiumDialogComponent: React.FC<FreemiumDialogComponentProps> = ({
   trigger,
   children,
   onClose,
+  startOnUpgradeModal,
 }) => {
   const { currentTheme } = useSelector(themeSelector);
-  const [activeModal, setActiveModal] = useState(ModalStage.ONE);
+  const [activeModal, setActiveModal] = useState(
+    startOnUpgradeModal === true ? ModalStage.TWO : ModalStage.ONE
```
iamnotreddy, line 35: if you're referencing a boolean in conditional logic the default case is true so you don't need to explicitly check if it's true

this can just be startOnUpgradeModal ? ModalStage.TWO : ModalStage.ONE

kwiedenhaupt: right! made the change!

PR #948, opened 2023-04-21
Add Fade Transition between the modals of the Onboarding Modal

kwiedenhaupt:
Purpose of the PR: 
Add the transition between the components in the onboarding modal.

1.  I modified the useTrail animation component I found online to include a fade transition and put it into the UI Animation folder.
2.  I wrapped the this component around the getModalContent function which switches between the modals and then used the activeModal variable as the key to trigger each transition.


iamnotreddy (changes requested):
I ran this branch locally and I don't think I'm seeing the animations unless I'm missing something.

Can you post a quick Loom of what I should be seeing

iamnotreddy (approved):
looks great! nice work

@LuckyHariharan - Kari has a pretty nice implementation of this fade animation, is this what you're kind of looking for? https://www.loom.com/share/0a1c5bb4acae4449a71b0a6011e0878e

If so there might be a way to reuse this component or at least try to follow its logic

PR #950, opened 2023-04-24
GraphQL implemetation for First Programming Quiz

jimip85:
This PR enhances the FirstProgrammingLanguageQuiz component by implementing mutation logic to efficiently handle user inputs such as their name, email, and quiz responses.

To achieve this, two mutations have been implemented:

(a) createQuizResponse mutation: to store the user input data (name and email) and generate a unique quizResponseId that is used to update the final quiz response object.
(b) updateQuizResponse mutation: to store the final quiz response object which includes the user-selected reasons, fields, interests, and quiz result.

Furthermore, the quizResponseId generated during the createQuizResponse mutation is crucial as it is used to update the final quiz response object before it is passed to the updateQuizResponse mutation for storage in the database.


LuckyHariharan (commented):
Excellent work implementing the upsert mutation fully! You code was easy for me to review and implement.

pages/resources/quizzes/firstLanguage/FirstProgrammingLanguageQuiz.tsx
```diff
@@ -1,3 +1,4 @@
+import { useMutation } from "@apollo/client/react/hooks/useMutation";
```
iamnotreddy, line 1: let's actually just name this file index.tsx and the folder firstProgrammingLanguage

you'll be able to navigate to the quiz like this: skillify.ca/resources/quizzes/firstProgrammingLanguage

If you're curious about how routing works in next JS: https://nextjs.org/docs/routing/dynamic-routes

pages/gql.tsx
```diff
@@ -0,0 +1,9 @@
+import { useQuery } from "@apollo/client";
```
iamnotreddy, line 1: this file shouldn't be in your PR

iamnotreddy (changes requested):
Great work on this PR! This feature isn't trivial by any means - you've tackled a ton of important react concepts the last few weeks. And I can tell by the way you're communicating that you're starting to get a good grasp / understanding of how React works. 

Left a couple comments on things to adjust but this is pretty much there

components/resources/quizzes/shared/BluePrint.tsx
```diff
@@ -1,14 +1,30 @@
...
 };
 
-const BluePrint = ({ onNextClick, onBackClick }: BluePrintProps) => {
+const BluePrint = ({
+  onNextClick,
+  onBackClick,
+  handleQuizResponseMutations,
+  quizViewState,
+}: BluePrintProps) => {
+  const result = getPreferredLanguageForQuizResults(
```
vithushan19, line 24: do you want this to be in the shared component?

components/resources/quizzes/shared/StartQuiz.tsx
```diff
@@ -22,19 +42,27 @@ const StartQuiz = ({ onNextClick, body, title }: StartQuizProps) => {
...
           ></input>
         </div>
         <Button
           backgroundColor="yellow"
           label="Start Quiz"
-          onClick={onNextClick}
+          disabled={userInput.name.length == 0 && userInput.email.length == 0}
+          onClick={() => {
+            onNextClick();
+            handleUserInputMutations(userInput);
```
vithushan19, line 64: this component seems to get more complicated just because you want to call this function here. What if you moved this function inside of of the onNextClick function that is passed in from the parent?

graphql/quizzes/upsertCodingLanguageQuiz.ts
```diff
@@ -0,0 +1,20 @@
+import { gql } from "@apollo/client";
+
+export const UPSERT_CODING_LANGUAGE_QUIZ_RESPONSE = gql`
+  mutation upsertCodingLanguageQuizResponse(
+    $objects: [coding_language_quiz_insert_input!]!
+  ) {
+    insert_coding_language_quiz(
+      objects: $objects
+      on_conflict: {
```
vithushan19, line 9: when does a conflict happen for this table?

graphql/quizzes/upsertCodingLanguageQuiz.ts
```diff
@@ -0,0 +1,20 @@
+import { gql } from "@apollo/client";
+
+export const UPSERT_CODING_LANGUAGE_QUIZ_RESPONSE = gql`
+  mutation upsertCodingLanguageQuizResponse(
+    $objects: [coding_language_quiz_insert_input!]!
```
vithushan19, line 5: are you allowing for multiple quiz responses to be inserted at once? Is there a way to simplify this query so you're only adding one row as needed. If you can't come up with a cleaner query, that's ok. It might not be possible

pages/resources/quizzes/firstProgrammingLanguage/index.tsx
```diff
@@ -35,11 +37,58 @@ const initializeQuizViewState = {
...
+    name: "",
+    email: "",
+  });
+
   const [triggerAnimation, setTriggerAnimation] = useState(true);
 
+  const [saveQuizResponse] = useMutation(UPSERT_CODING_LANGUAGE_QUIZ_RESPONSE, {
+    onCompleted: (data) => {
+      if (!quizResponseId) {
+        setQuizResponseId(
```
vithushan19, line 54: why is it important to save this? how do you reuse this id later?

iamnotreddy: @vithushan19 - can chime in here - the idea is that we write to the table when the user saves their information so we don't lose name / email, save the row id of what's inserted in state, and modify the same row through the same query when we get all the quiz responses

graphql/quizzes/upsertCodingLanguageQuiz.ts
jimip85: Separated the logic. We now have an insert and update query!

graphql/quizzes/upsertCodingLanguageQuiz.ts
jimip85: I was using this to handle duplicate rows. But, the new implementation completely removes this.

components/resources/quizzes/shared/StartQuiz.tsx
jimip85: Done!

components/resources/quizzes/shared/BluePrint.tsx
```diff
@@ -1,4 +1,3 @@
-import React from "react";
```
vithushan19, line 1: don't think we want to include this change

components/resources/quizzes/shared/BluePrint.tsx
```diff
@@ -40,7 +39,9 @@ const BluePrint = ({ onNextClick, onBackClick }: BluePrintProps) => {
       <div className="grid py-6 place-items-center">
         <Button
           label="View results"
-          onClick={onNextClick}
+          onClick={() => {
```
vithushan19, line 42: there's no need for this file to be changed at all in your PR. I would just remove the formatting change if possible to make your PR simpler

graphql/quizzes/insertCodingLanguageQuiz.ts
```diff
@@ -0,0 +1,11 @@
+import { gql } from "@apollo/client";
+
+export const INSERT_CODING_LANGUAGE_QUIZ_RESPONSE = gql`
+mutation insertCodingLanguageQuizResponse($name: String!, $email: String!) {
```
vithushan19, line 4: did you test this mutation in hasura playground?

graphql/quizzes/updateCodingLanguageQuiz.ts
```diff
@@ -0,0 +1,23 @@
+import { gql } from "@apollo/client";
+
+export const UPDATE_CODING_LANGUAGE_QUIZ_RESPONSE = gql`
```
vithushan19, line 3: nice!

graphql/quizzes/updateCodingLanguageQuiz.ts
```diff
@@ -0,0 +1,23 @@
...
+mutation updateCodingLanguageQuizResponse(
+    $id: Int!,
+    $reasons: jsonb,
+    $fields: jsonb,
+    $interests: jsonb,
+    $result: String
+  ) {
+    update_coding_language_quiz(
+      where: {id: {_eq: $id}},
+      _set: {
```
vithushan19, line 13: nice!

pages/resources/quizzes/firstProgrammingLanguage/index.tsx
```diff
@@ -35,14 +42,59 @@ const initializeQuizViewState = {
...
+          setQuizResponseId(
+            parseInt(data.insert_coding_language_quiz.returning[0].id)
+          );
+        }
+      },
+    }
+  );
+
+  //Update Mutation - Store the rest of the columns via id.
+  const [finalQuizResponse] = useMutation(UPDATE_CODING_LANGUAGE_QUIZ_RESPONSE);
```
vithushan19, line 70: I would rename this to updateQuizResponse. 

finalQuizResponse is a noun, which makes me think that the `useMutation` hook is returning an object, but it's not it's returning a function that will do something, so we should use a verb instead.

pages/resources/quizzes/firstProgrammingLanguage/index.tsx
```diff
@@ -35,14 +42,59 @@ const initializeQuizViewState = {
...
+    result: getPreferredLanguageForQuizResults(
+      computeLanguageScore(quizViewState)
+    ),
+  };
 
   const handleNextClick = () => {
     setTriggerAnimation(false);
 
+    //triggering mutations via onNextClick
+    if (stage == Stage.START) {
```
vithushan19, line 92: nice

vithushan19 (approved):
looks good to me. just had a minor comment about naming. great work with the graphql

graphql/quizzes/insertCodingLanguageQuiz.ts
jimip85: yes, and it works!

iamnotreddy (approved):
🚀

PR #952, opened 2023-04-25
add sign up flow to plans page

iamnotreddy:
Implemented sign up with google on the plans page. I also reworked the component and its corresponding types to be a little cleaner. When the user clicks sign up, they go to the google auth flow. Clicking on Book Call takes them to the calendly appointment page.

image

A couple other things:
- there was something weird going on with the default color on the Button component not showing orange unless you hover over it... couldn't figure out exactly what was going on, but the other colors work. I thought white looked a little cleaner anyways.
- I was going to throw a google logo in the button but it was getting a little tricky with passing the right handler logic. Can definitely work this in, or add some sort of caption below the button to let the user know we only support signing in with gmail
- We should definitely test the full flow to make sure a new user routes right to the studentPortal page and sees the new modals

### DIFF
--- a/components/studentPortal/freemium/PlansCard.tsx
+++ b/components/studentPortal/freemium/PlansCard.tsx
@@ -1,34 +1,38 @@
-import React from "react";
+import { useRouter } from "next/router";
+import { useAuth } from "../../../lib/authContext";
+import { PlanCard } from "../../../pages/plans";
 import { Button } from "../../ui/Button";
-import PlanRow, { PlanRowProps } from "./PlanRow";
+import PlanRow from "./PlanRow";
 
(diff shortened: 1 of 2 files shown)


vithushan19 (approved):

kwiedenhaupt:
Just if you wanted to fix it or in the future -- the button color doesn't work I think because it has the color bg-brandPrimary which was set up in the globals.css file.  

If you use the   const { currentTheme } = useSelector(themeSelector); and set the currentTheme in the component you want it to apply to, it should work...But I agree with you.  I think white actually looks better!!

thank you for reworking the types, etc and making it cleaner!

PR #958, opened 2023-04-27
Add Webhook to Freemium

kwiedenhaupt:
I only had a few minutes and I'll need some help to tweak it, but the webhook does work. 

Can @vithushan19 or @iamnotreddy help as to where this hook should be placed in the signup flow? 

I think it's a great idea to do this!


pages/api/studentPortal/freemium/helpers.ts
```diff
@@ -75,3 +75,17 @@ function transformBackendUnit(unit: Unit): Unit {
     }),
   };
 }
+
+export const sendSlackNotification = async () => {
+  try {
+    const response = await fetch(
+      `/api/slack/${"new free trial registration"}`
```
vithushan19, line 82: can we feed in the users email here as well?

vithushan19 (changes requested):

iamnotreddy (commented):
I pushed a commit to add the User type so you can see more easily see what's accessible in the user object. There is a name, email, creation date, among other useful things which will help with Vithushan's ask.

In terms of where to put the function - I don't think you should place it right after signIn() because it would run even if they don't successfully create an account.

I see two potential options:
- do a check in the useEffect to see whether creation time from the user object is relatively recently from now (maybe difference < 1 hour?)... this technically isn't perfect because we could get multiple notifications if the user keeps hitting that flow (browser unexpectedly refreshes, accidentally navigates to the same page, etc.)
- Call the slack notif function when the upsert query is run. This occurs inside the userSync function inside authContext.js file. I think you would have to modify the query mechanics to know whether it's upserting (if the user exists) or inserting (an actual sign up) though. 

You haven't worked with the auth files so this might be kind of complicated so approach 2 might be more complicated. @vithushan19 - do you think approach one is acceptable (or is there another better way to do this?)

components/studentPortal/layout/useLastSeenModal.tsx
```diff
@@ -17,6 +22,7 @@ export const useLastSeenModal = (
   const [updateLastSeenModal] = useMutation(UPSERT_LAST_SEEN_MODAL);
   const [showOnboardingModal, setShowOnboardingModal] = useState(false);
   const [showExitModal, setShowExitModal] = useState(false);
+  const { email } = useSelector(profileSelector);
```
vithushan19, line 25: this could be 
`const { user } = useAuth()`
`const email = user.email`

and you could get rid of all your redux stuff. There's no need for your redux changes, since we already have the users email inside the useAuth hook

vithushan19:
@raveenreddy0 the best place to call the notification is when the new user is created inside of user sync, but since it would require a refactor I am ok with getting duplicate notifications. 

Maybe even trigger it anytime the user sees the freemium modal. If they open the modal we can call the notification. I don't expect to get overwhelmed with duplicate notifications, and if we do, then we can just refactor the user sync query.

Ideally we would refactor the query mechanics and just send a notification whenever a new user was created in the table regardless if they came through the freemium feature or not

components/studentPortal/freemium/FreemiumDialogueComponent.tsx
```diff
@@ -20,8 +20,8 @@ export enum ModalStage {
 export interface FreemiumDialogComponentProps {
   trigger: boolean;
   triggerTitle?: string;
-  onClose: () => void;
-  startOnUpgradeModal: boolean;
+  onClose?: () => void;
```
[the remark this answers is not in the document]
kwiedenhaupt: That only has to do with having the dialog box remain open when it is called from the tooltip so that when you move your cursor out of the tooltip it doesn't automatically close without you seeing the modal.  It is not needed for the first automatic onboarding.

components/studentPortal/layout/Layout.tsx
```diff
@@ -113,7 +108,8 @@ export const Layout: React.FC = ({ children }) => {
       >
         <Sidebar />
       </div>
-      {showOnboardingModal && <FreemiumDialogComponent />}
+      {showOnboardingModal && <FreemiumDialogComponent trigger={false} />}
```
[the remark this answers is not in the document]
kwiedenhaupt: that controls whether or not the modal needs you to click on something to trigger its opening or for it to come up automatically.  we need that because it needs to come up automatically on the first load in the layout, but you have to click on the icon in the tooltip for it to show in the tooltip on the sidebar.

components/studentPortal/layout/useLastSeenModal.tsx
kwiedenhaupt: great idea! so much cleaner.

kwiedenhaupt:
There were many notifications in the slack channel because I tested it many times lol

vithushan19 (approved):
nice work!

PR #959, opened 2023-04-27
Quiz Styling Fixes

LuckyHariharan:
These changes address only styling issues. Refactoring of code will be another PR. Please feel free to share your thoughts.
- Responsive design styling for the BluePrint.tsx component
image

- Refactor QuizTransition.tsx to decrease duration to 200 and remove initial horizontal animation, which results in a snappier animation. no refactoring needed for quiz index files for this to work. havent touched setTimeout in this PR.
animationdemo
- Edubackground.tsx styling change so that Institution and degree line up with Experience with coding
image
image
image

### DIFF
--- a/components/resources/quizzes/careerQuiz/CareerResults.tsx
+++ b/components/resources/quizzes/careerQuiz/CareerResults.tsx
@@ -23,7 +23,7 @@ const CareerResults = ({ onBackClick, quizViewState }: CareerResultsProps) => {
         <img
           src={quizResult && quizResult.src}
           alt={quizResult && quizResult.alt}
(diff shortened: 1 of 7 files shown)


vithushan19 (approved):

iamnotreddy (changes requested):
Mobile styling looks great but we need to consider desktop as well. Can we reduce the width of the progress bar and input boxes to match the width quiz question template. Think it should just be a case of modifying your breakpoints

image

iamnotreddy (approved):
changed to max-w-xl so the input and title line up better on the education results page

approved!

PR #961, opened 2023-04-27
Implemented MaxSelection & ProgressBar into firstProgrammingQuiz

jimip85:
In this PR, the following changes were made:

- Added a useEffect hook to update the progress bar dynamically based on the current question number and the total number of questions.
- Added a useEffect hook to monitor the number of selected options on each page.
Updated the quiz component to display the progress bar and the selected option count.


iamnotreddy (approved):
🚀

PR #965, opened 2023-04-29
Freemium UI Fixes per Grace's comments

kwiedenhaupt:
Purpose of PR: 
To fix UI issues that Grace pointed out.

image
image
image

COMMENTS:  I noticed that I had been developing with 67 to 80 percent screen zooming and when I moved to 100, then I hit the problems Grace was having because of the top nav bar was covering the modal.  

COMMENT 1: could we move the top text down a bit? It looks cut off
I need everyone's opinion on this, but I adjusted some spacing and the main change I made was changing where the modal would appear on the screen, using the translate and top in this div and the height to make it overall shorter: 
<div
            className={`fixed h-[450px] w-[300px] **md:h-[550px]** md:w-[900px] md:p-10 px-4 transform -translate-x-1/2 md:-**translate-y-3/4** -translate-y-1/2 ${
              activeModal === ModalStage.TWO ? "bg-murkrow" : "bg-white"
            } rounded-lg left-1/2 top-1/2 **md:top-3/4**`}
          >

COMMENT 2 was a small fix. 
COMMENT 3: also kinda nitpicky but I noticed that the 3 dots in the nav bar on the third modal isn't in the same spot as the first two modals, it shifts to the right
For this, I adjusted the SVG for the "Done" image.  The spacing problem is due to it doesn't have the > sign after it like NEXT> in the svg, so the SVG itself is a bit skinnier and that shifts the dots.  I changed the SVG to be the same size, but then the word "DONE" was too large.  I then altered that, and had to move it slightly down as well.  Let me know if you think this is an improvement! or we should leave it.  The height could be slightly off.

Another option would be to put an invisible ">" in the figma design next to DONE and create the SVG From that.


vithushan19 (approved):
lgtm!

vithushan19:
I think your solution is good, but you should have been able to manage without changing the done image and extra padding. You should have been able to add that in code. Either way, it's not a big deal.

PR #966, opened 2023-05-02
Address Feedback for Career Quiz

LuckyHariharan:
Refactored Career Quiz to address the following feedback:
- Going back to the education page now shows previously entered responses
- Added consistent `px-4` padding for all text and select inputs
- Change copy on the results page to remove “Click below to learn more”

The back-end seems to be working fine. I've not run into any more problems and reviewing the code has not revealed any cause for issue.


vithushan19 (approved):
nice work!

PR #978, opened 2023-07-17
Create file for interview course

Zelaney02:
This PR creates a new file for the interview course.


components/landingPage/Hero.tsx
```diff
@@ -1,4 +1,5 @@
-import Link from "next/link";
+import React from 'react';
```
vithushan19, line 1: imported React to fix lint warnings

PR #981, opened 2023-07-28
Add NextJS API route to both quizzes - CareerQuiz, FirstProgrammingLanguage

LuckyHariharan:
Added API endpoints to both careerQuiz.ts and firstProgrammingLanguage.ts

@jimip85 please let me know if ive done this incorrectly.


iamnotreddy (approved):
only one comment - the way you have it named right now, you'll need to hit /api/studentPortal/quizzes/careerQuiz/careerQuiz which seems unnecessary unless you're planning on having multiple quizzes in the careerQuiz folder.. same goes for the other quiz

PR #1000, opened 2023-10-31
Fetch user profile pictures from AWS S3 bucket

vithushan19:
This PR fetches profile images from an S3 bucket on AWS. We use the AWS Typescript SDK to interact with the S3 Bucket easily and programmatically.

This PR reads files from the bucket. A follow up PR will write or upload files to the bucket.

### DIFF
--- a/components/studentPortal/layout/Sidebar.tsx
+++ b/components/studentPortal/layout/Sidebar.tsx
@@ -1,5 +1,4 @@
 import { useQuery } from "@apollo/client/react";
-import { getSignedUrl } from "@aws-sdk/s3-request-presigner";
 import Link from "next/link";
 import { useRouter } from "next/router";
 import React, { useEffect, useState } from "react";
@@ -9,6 +8,7 @@ import {
   FetchRoleData,
 } from "../../../graphql/studentPortal/users/fetchUserRole";
 import { useAuth } from "../../../lib/authContext";
+import { fetchProfilePicture } from "../../../pages/profile/profilePicturesClient";
 import {
   profileSelector,
   setCreatedAt,
@@ -24,11 +24,6 @@ import SkillifyCommandPalette from "./CommandPalette";
 import SidebarItem, { SidebarItemProps } from "./SidebarItem";
 
 // import aws s3
-import {
-  GetObjectCommand,
-  ListObjectsV2Command,
-  S3Client,
-} from "@aws-sdk/client-s3";
 
 export const Sidebar: React.FC = () => {
   const { goalApproaching } = useSelector(activePageSelector);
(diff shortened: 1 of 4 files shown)


components/studentPortal/layout/Sidebar.tsx
```diff
@@ -74,6 +79,18 @@ export const Sidebar: React.FC = () => {
     }
   }, [router.pathname]);
 
+  useEffect(() => {
+    if (user) {
+      fetchProfilePicture(user.uid).then((imageUrl) => {
```
vithushan19, line 84: as a frontend developer i should be able to fetch the profile picture for the current logged in user easily and not have to worry about where the image is coming from. 

if we had more complex backend systems and needed to migrate from AWS to Google Cloud, this front end code shouldn't need to change.

pages/profile/profilePicturesClient.ts
```diff
@@ -0,0 +1,46 @@
...
+      secretAccessKey: process.env.NEXT_PUBLIC_AWS_SECRET_ACCESS_KEY,
+    },
+  });
+
+  const params = {
+    Bucket: "student-profile-pictures",
+    Prefix: `${userId}/`,
+  };
+  return s3
+    .send(new ListObjectsV2Command(params))
```
vithushan19, line 25: to interact with an aws s3 bucket you need to understand the possible commands that you can run. the list of commands is availble at the s3 documentation page

https://docs.aws.amazon.com/AmazonS3/latest/API/API_Operations_Amazon_Simple_Storage_Service.html

pages/profile/profilePicturesClient.ts
```diff
@@ -0,0 +1,46 @@
...
+    .then(async (res) => {
+      if (res.Contents && res.Contents.length > 0) {
+        const firstImage = res.Contents.filter((x) => !x.Key.endsWith("/"))[0]
+          .Key;
+        const imageUrl = await getSignedUrl(
```
vithushan19, line 30: this signedUrl expires so that Amazon has more control and restrictions over how our data is accessed. I can't save this url in my database for later, I always have to ask amazon for a new image url to display

components/studentPortal/layout/Sidebar.tsx
jimip85: Awesome! Ensuring the front-end code remains agnostic to back-end changes is a brilliant way to guarantee our system's flexibility in the long run.

pages/api/studentPortal/profile/profilePicturesClient.ts
jimip85, line 1: This is very clear and easy to understand. I'm definitely going to implement this into Promptify as well.

jimip85 (approved):

PR #1003, opened 2023-11-02
Create Message Feed

Gawshan987:
Here I've created the side bar for the messages that will be displayed after the user and entered their message. You would have to refresh the page to see it update on the side bar.
image


vithushan19 (approved):
looks good to me.

PR #1009, opened 2023-12-05
Create Virtual Hangout (1 Player Version)

LuckyHariharan:
1 Player MVP -- Skillify Virtual Hangout:
image

- movement using WASD, arrow keys, or clicking/tapping
- text chat -- toggle with Q
- randomized sprites

virtualdemo


vithushan19 (commented):
Do we need these changes to the package json if we're just using an iframe?

LuckyHariharan:
> Do we need these changes to the package json if we're just using an iframe?

I tried to remove all changes that I dont need, but I think i need axios because im missing it for some reason.

PR #1021, opened 2025-01-09
Interview Tracker

LuckyHariharan:
## What This PR Does
- Adds an **Interview Tracker** section to my student profile, Re-using the `OfferTable` Component
- Displays interview data in a table format:
  - **Columns:** Company, Role, Location, Timeline, Result, Asked Task.
- Ensures that the tracker is only visible when the `userId` matches mine.

I also made **two changes** to the content of my data to make it more relevant to me:
- **Replaced `opportunitySource` with `askedTask`:** Instead of showing where I found the opportunity since all are from LinkedIn/Website Applications, I now display what I was asked to do in the interview (e.g., "Livecoding - Tic Tac Toe App").
- **Replaced `timeSpent` with `timeline`:** Removed the time spent interviewing and replaced it with when the interview occurred (e.g., "March 2024") for better chronological tracking.

interviewsection

## Why These Changes Were Made
- **Relevance:** Tracking tasks and timelines is more meaningful and actionable than opportunity source or time spent.
- **Personalization:** The tracker is specific to my profile and not shown for other users.
image


pages/profile/[userId].tsx
```diff
@@ -230,6 +233,15 @@ export default function InternalProfile({
         <AchievementComponent userId={userId} />
       ),
     },
+    {
+      shouldShow:  userId == "R7nzMKiRewgJuLm54eQ1KdHV3g82",
```
Comment (LuckyHariharan) [inline pages/profile/[userId].tsx:237]: @vithushan19  what is the best form when referencing my userId? Or is this fine how it is.

LuckyHariharan (commented):
Ready for peer review by @jimip85  @Curtoista

components/resources/jobTracker/JobTrackerComponent.tsx
```diff
@@ -582,7 +583,9 @@ function OfferTable({ isCAD, year }) {
     },
   ];
 
-  const offers = year === 2023 ? offers2023 : offers2024;
+  // If no data is supplied, use 2023 or 2024 data below
+  // Otherwise use the supplied data for the profile
```
vithushan19, line 587: Nice

pages/api/studentPortal/interviews/LuckyInterviewData.ts
```diff
@@ -0,0 +1,138 @@
+import { Offer } from "../../../../components/resources/jobTracker/JobTrackerComponent";
+// Lucky H Interview Data 
+export const LuckyInterviewData: Offer[]=[
```
vithushan19, line 3: Awesome!

pages/api/studentPortal/interviews/LuckyInterviewData.ts
```diff
@@ -0,0 +1,138 @@
...
+    {
+      "Company": "Robinhood",
+      "companyUrl": "https://www.robinhood.com/",
+      "Location": "Toronto",
+      "Role": "Mobile Developer",
+      "Time Spent": "March 2024",
+      "timeSpentHours": 10,
+      "Result": "Failed First Technical Round",
+      "Base": 115000,
+      "opportunitySource": "Livecoding - List of instruments with filtering",
```
vithushan19, line 13: I think we should keep opportunity score and time spent. I agree adding what you were asked to do and having the timeline is helpful but I would add those as additional properties and just optionally render them on your profile view. 

It's confusing to see the property name not match the value in this data. 

Add your custom properties instead of replacing existing ones. Your custom properties can be optional so you don't affect the other offer table components

pages/profile/[userId].tsx
Comment (vithushan19) [reply pages/profile/[userId].tsx:237]: At the top of this profile component you should create a const interviewData = getInterviewData(userId)

getInterviewData should be a function defined in the api folder probably in studentPortal/interviews/

In the api function you can do the check against your specific userID and then return your hardcoded data if the id matches. 

If the id doesn't match then return null. 

This way shouldShow can check if the interviewData is null or not.

pages/profile/[userId].tsx
```diff
@@ -230,6 +233,15 @@ export default function InternalProfile({
...
+      title: `Interview Tracking`,
+      hasProgress: true,
+      value: (userBadgeCount * 100) / totalBadgeCount,
+      component: typeof userId == "string" && (
+        <OfferTable  isCAD={isCAD} year={year} data={LuckyInterviewData}/>
```
Comment (vithushan19) [inline pages/profile/[userId].tsx:242]: Instead of calling your hardcoded data here, create a const like I suggested earlier and use that here. It could potentially be null if the user id doesn't match

vithushan19 (commented):
You are on the right track! Very close. 

Great job putting your data together. 

Don't change existing properties in the table. Add new ones and make it optional

pages/api/studentPortal/interviews/LuckyInterviewData.ts
LuckyHariharan: Right makes sense, sounds good to me.

pages/profile/[userId].tsx
Comment (LuckyHariharan) [reply pages/profile/[userId].tsx:242]: will do, sounds good

vithushan19 (approved):
Going to approve this to unblock you, since jimi would benefit from these changes being merged in. Please address my earlier comments before merging.

LuckyHariharan:
> Going to approve this to unblock you, since jimi would benefit from these changes being merged in. Please address my earlier comments before merging.

Will address these changes and merge today/this weekend. Thanks again for your feedback!

PR #1035, opened 2025-01-31
refactor: remove ai engineer components in favour of the re-usable one

vithushan19:
We now have a re-usable job component so we can delete these hard coded ones. 

this PR shows how to delete the hard-coded AI engineer components. you should be able to follow this patter for other jobs.

the re-usable page uses a slug and can be found here: https://github.com/skillify-ca/skillify-web/blob/vn-remove-ai-engineer-components/pages/resources/jobExplorer/%5Bslug%5D.tsx#L42

but you shouldn't have to touch that page. You just need to update the fetch function to read the data for any new jobs, and it should all just work.


LuckyHariharan:
Thanks for doing this. Took a look and will chat with @Curtoista about how we want to split this.

pages/api/resources/jobExplorer/aiEngineer/data.ts
```diff
@@ -77,3 +77,11 @@ export const project = {
   name: "AI Engineer Project",
   link: "aiEngineerProject",
 };
+
```
LuckyHariharan, line 80: cool, will update the remaining projects to make sure we follow this pattern

LuckyHariharan:
Sounds good to me, approved. Thanks for doing this!

LuckyHariharan (approved):

